PR #6, opened 2022-02-07
IRQ-based VirtIOBlk access.

wyfcyx:
Now VirtIOBlk device can be accessed in a non-blocking way based on IRQ, and we keep the interface compatible with the ecosystem which relies on this library since we did not change or remove any existing interface.

We only added 5 methods:
* `VirtIOBlk::read_block_nb`
* `VirtIOBlk::write_block_nb`
* `VirtIOBlk::pop_used`
* `VirtIOBlk::virt_queue_size`
* `VirtQueue::size`(only called by `VirtIOBlk::virt_queue_size`)

Their usage has been included in the code documentation.

## How to use them
1. Initialization: Call `virt_queue_size` to get the number of channels the caller should monitor on. Every descriptor in the descriptor table is called a channel. For example:
```rust
// condvars: BTreeMap<u16, Condvar>
let mut condvars = BTreeMap::new();
let channels = virtio_blk.exclusive_access().virt_queue_size();
for i in 0..channels {
    let condvar = Condvar::new(); 
    condvars.insert(i, condvar);
}
```
2. Read/write a block: Call `read_block_nb` or `write_block_nb` to get the token of the request immediately, then block on the channel whose identifier is the same as the token. After the thread is woken up, we can read `resp` which stores the status of the request to check if the request was successfully completed.
```rust
let mut blk = self.virtio_blk.exclusive_access();
let mut resp = 0xffu8;
let token = blk.write_block_nb(block_id, buf, &mut resp).unwrap();
drop(blk);
self.condvars.get(&token).unwrap().wait();
assert_eq!(resp, 0x0, "Error when reading VirtIOBlk");
```
3. Handling interrupts: When an interrupt is received, the user calls `pop_used` to fetch all of the completed requests' token from the used ring, and then wake up the threads that block on the corresponding channels. For example:
```rust
fn handle_irq(&self) {
    let mut blk = self.virtio_blk.exclusive_access(); 
    while let Ok(token) = blk.pop_used() {
        self.condvars.get(&token).unwrap().signal();
    }
}
```
## Publicity of `BlkResp`
Now the user should provide a variable `resp` of type `u8` to store the status of the request. Should we make [`BlkResp`](https://github.com/rcore-os/virtio-drivers/blob/master/src/blk.rs#L122) public so that we can use it instead of `u8` to enhance the type safety? I have no idea yet.

### DIFF
--- a/src/blk.rs
+++ b/src/blk.rs
@@ -77,7 +77,7 @@ impl VirtIOBlk<'_> {
     ///
     /// * `block_id` - The identifier of the block to read.
     /// * `buf` - The buffer in the memory which the block is read into.
-    /// * `resp` - A mutable reference to a variable provided by the caller 
+    /// * `resp` - A mutable reference to a variable provided by the caller
     ///   which contains the status of the requests. The caller can safely
     ///   read the variable only after the request is ready.
     ///
@@ -89,17 +89,19 @@ impl VirtIOBlk<'_> {
     ///
     /// After the request is ready, `resp` will be updated and the caller can get the
     /// status of the request(e.g. succeed or failed) through it. However, the caller
-    /// **must not** spin on `resp` to wait for it to change. A safe way is to read it 
+    /// **must not** spin on `resp` to wait for it to change. A safe way is to read it
     /// after the same token as this method returns is fetched through [VirtIOBlk::pop_used()],
     /// which means that the request has been ready.
     pub fn read_block_nb(&mut self, block_id: usize, buf: &mut [u8], resp: &mut u8) -> Result<u16> {
-        assert_eq!(buf.len(), BLK_SIZE); 
+        assert_eq!(buf.len(), BLK_SIZE);
         let req = BlkReq {
             type_: ReqType::In,
             reserved: 0,
             sector: block_id as u64,
         };
-        let token = self.queue.add(&[req.as_buf()], &[buf, core::slice::from_mut(resp)])?;
+        let token = self
+            .queue
+            .add(&[req.as_buf()], &[buf, core::slice::from_mut(resp)])?;
         self.header.notify(0);
         Ok(token)
     }
@@ -131,7 +133,7 @@ impl VirtIOBlk<'_> {
     ///
     /// * `block_id` - The identifier of the block to write.
     /// * `buf` - The buffer in the memory containing the data to write to the block.
-    /// * `resp` - A mutable reference to a variable provided by the caller 
+    /// * `resp` - A mutable reference to a variable provided by the caller
     ///   which contains the status of the requests. The caller can safely
     ///   read the variable only after the request is ready.
     ///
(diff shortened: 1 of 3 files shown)


wangrunji0408 (approved):
LGTM!

For the issue "Publicity of `BlkResp`", I think we can make an opaque type `Token(u16)` instead.

Besides, note that there are potential unsafe behaviors in the non-blocking functions:
```rust
fn read_block_nb(..., buf: &mut [u8]);
fn write_block_nb(..., buf: &[u8]);
```

The borrow on `buf` is released after the function returns. However it is still referenced by the underlying block device until IO completed. So maybe we should mark these functions with `unsafe`.

wyfcyx:
I agree with you that `read_block_nb` and `write_block_nb` should be marked with `unsafe`.
As for `BlkResp`, it is not a token of type `u16`. It is of type `u8` indicating the completion status of the request, and the virtio-blk device writes to it after a request is completed. To make it more clear, see the code below:
```rust
#[repr(C)]
#[derive(Debug)]
struct BlkResp {
    status: RespStatus,
}
#[repr(u8)]
#[derive(Debug, Eq, PartialEq)]
enum RespStatus {
    Ok = 0,
    IoErr = 1,
    Unsupported = 2,
    _NotReady = 3,
}
```
My suggestion is making `BlkResp` public thus the signature of `read_block_nb` and `write_block_nb` can be changed to
```rust
fn read_block_nb(&mut self, block_id: usize, buf: &mut[u8], resp: &mut BlkResp);
fn write_block_nb(&mut self, block_id:usize, buf: &[u8], resp: &mut BlkResp);
```
respectively by substituting `&mut u8` with `&mut BlkResp`. After that, we can check if the request successfully completed using `resp.status == RespStatus::Ok`, eliminating hardcoded constants from the code.
When it comes to the token, I think it is not necessary to wrap `u16` into a `Token(16)` since it just make things more complex.

wangrunji0408:
Oh I see. No problem.

wyfcyx:
> Oh I see. No problem.

Currently in `lib.rs` I added:
```rust
pub use self::blk::{BlkResp, RespStatus, VirtIOBlk};
```
I am not sure if this is a good idea. If you think it is ok, then I would like to merge this into master.

src/blk.rs
```diff
@@ -92,16 +92,26 @@ impl VirtIOBlk<'_> {
...
+    ///
+    /// `buf` is still borrowed by the underlying virtio block device even if this
+    /// method returns. Thus, it is the caller's responsibility to guarantee that
+    /// `buf` is not accessed before the request is completed in order to avoid
+    /// data races.
+    pub unsafe fn read_block_nb(
+        &mut self,
+        block_id: usize,
+        buf: &mut [u8],
+        resp: &mut BlkResp,
```
wangrunji0408, line 106: How does user initialize a `BlkResp` before calling it? Seems a `new` or `default` method is missing?

wyfcyx: Now I just use `BlkResp::default()`, which is also used in `read_block` and `write_block`.

PR #11, opened 2022-08-18
Use a safer approach to the HAL interface

qwandor:
Rather than rely on `extern "C"` functions, this makes the HAL a trait that must be implemented and provided as a type parameter. This lets it be properly type checked and removes a bunch of `unsafe` blocks.


wangrunji0408 (approved):
Looks cool!

Could you rebase this branch so it can get merged? Thanks

PR #12, opened 2022-08-23
Add tests for VirtQueue

qwandor:
I'd like to add tests for the device implementations too, but that's going to require some refactoring to allow the transport to be faked somehow.


wangrunji0408 (approved):
LGTM. Thanks!

PR #14, opened 2022-08-31
Factor out trait for transport, and add support for modern MMIO transport

qwandor:
This PR makes a bunch of changes around the VirtIO transport implementation.

* I've added a new `Transport` trait. This currently only has one implementation (for the MMIO transport) but I plan to add a PCI implementation in a future PR.
* I've added support for the modern MMIO transport (version 2) in addition to the legacy one (version 1). Initially I did this as two separate implementations of `Transport`, but this turned out to be rather cumbersome to use so I've gone for a single implementation that stores the version and checks it as needed at runtime. This is only relevant during initialisation, so shouldn't have a significant runtime cost.
* As part of the above, I made the new `MmioTransport` wrap a `NonNull` pointer to the `VirtIOHeader` rather than having the caller create a mutable reference to it. I'm still not entirely convinced that the way this crate uses references is sound, and will think about that some more, but I think this is a step in the right direction at least.

### DIFF
--- a/examples/riscv/Makefile
+++ b/examples/riscv/Makefile
@@ -36,7 +36,7 @@ header:
 clean:
 	cargo clean
 
-qemu: kernel $(img)
+qemu-legacy: kernel $(img)
 	qemu-system-$(arch) \
 		-machine virt \
 		-serial mon:stdio \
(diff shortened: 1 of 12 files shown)


qwandor:
Hi, any thoughts on this? I have some other changes I'd like to send that build on it, but I can't easily create dependent PRs without being a member of the project.

DeathWish5:
Thanks. Sorry for our late reply.

PR #16, opened 2022-09-14
Use actual GPU resolution in example rather than hardcoding.

qwandor:
Different versions of QEMU seem to have different default resolutions, so it was crashing due to an array access out of bounds.


wangrunji0408 (approved):
LGTM!

PR #18, opened 2022-09-23
Fix soundness issues with MMIO and shared memory

qwandor:
The way the `volatile` crate handles MMIO regions is not sound, because it creates references to memory which shouldn't be considered dereferenceable. There is an attempt to fix this in https://github.com/rust-osdev/volatile/pull/22 (and lots of discussion), but that PR has been open since May 2021 and doesn't seem near being submitted. Instead, I've implemented the functionality we need in a new `volatile` module here using the `addr_of!` and `addr_of_mut!` macros, keeping all MMIO regions as pointers and never creating references.

This works for the MMIO transport and config regions. For the Virtqueue, I don't think we need volatile access at all because it's just shared memory, not MMIO. We still shouldn't be using references, however, as DMA to these shared regions by the device violates Rust's aliasing rules, so I've changed the implementation to use pointers instead.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,5 @@ description = "VirtIO guest drivers."
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-volatile = "0.3"
 log = "0.4"
 bitflags = "1.3"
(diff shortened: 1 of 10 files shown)


wangrunji0408 (approved):
LGTM. I didn't realize this problem before. Thanks for your work!

PR #25, opened 2022-11-03
Use fdt crate rather than device_tree in riscv example.

qwandor:
It has a nicer API, and has a released crate rather than a git repository.


wangrunji0408 (approved):
Looks cool!

PR #30, opened 2022-11-10
Reset queues when device is dropped.

qwandor:
Otherwise the device might continue accessing them, despite the DMA memory having been deallocated, i.e. a use-after-free.

### DIFF
--- a/src/transport/mmio.rs
+++ b/src/transport/mmio.rs
@@ -418,6 +418,34 @@ impl Transport for MmioTransport {
         }
     }
 
+    fn queue_unset(&mut self, queue: u16) {
+        match self.version {
+            MmioVersion::Legacy => {
+                // Safe because self.header points to a valid VirtIO MMIO region.
+                unsafe {
+                    volwrite!(self.header, queue_sel, queue.into());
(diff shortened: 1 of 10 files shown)


src/transport/mmio.rs
```diff
@@ -418,6 +418,34 @@ impl Transport for MmioTransport {
...
+            }
+            MmioVersion::Modern => {
+                // Safe because self.header points to a valid VirtIO MMIO region.
+                unsafe {
+                    volwrite!(self.header, queue_sel, queue.into());
+                    volwrite!(self.header, queue_ready, 0);
+                    volwrite!(self.header, queue_num, 0);
+                    volwrite!(self.header, queue_desc_low, 0);
+                    volwrite!(self.header, queue_desc_high, 0);
+                    volwrite!(self.header, queue_driver_low, 9);
```
wangrunji0408, line 440: Is the 9 a typo?

qwandor: Whoops, it is indeed. Thanks for spotting that; fixed.

wangrunji0408 (approved):

PR #31, opened 2022-11-15
Fix soundness issues with VirtQueue

qwandor:
`Queue::add` must be marked as unsafe, because the caller must ensure that the input and output buffers remain valid until device has finished using them. I've added a helper method `Queue::add_notify_wait_pop` to handle the blocking case safely. This doesn't return until the matching token has been popped from the queue.


equation314 (approved):
LGTM

PR #33, opened 2022-11-17
Test more things in CI

qwandor:
* Try building and running tests with no features as well as with all features.
* Use actions-rs for Clippy, so we get source annotations from it.
* Build and run aarch64 and riscv examples in CI.


wangrunji0408 (approved):
LGTM!

PR #39, opened 2022-12-20
Improve VirtIOConsole

qwandor:
* Added a method to get information such as the number of rows and columns.
* Keep track of the token of the outstanding receive request, and ensure there is never more than one.
* Check whether the request has completed at the start of `recv` as well as in `ack_interrupt`, in case the client is not using interrupts.
* Added some more documentation.
* Added a test in the aarch64 example.


wangrunji0408 (approved):
LGTM! Thanks!

Issue #7, opened 2022-04-13
a bug in blk::read/write_block_nb


wyfcyx:
In these methods, we allocate a `BlkReq` instance on the stackframe. The memory block which the `BlkReq` instance resides in will be reused after the methods return, and there is a possibility that the virtio device is also accessing it at the same time, leading to a data race. Mention that the blocking version of there methods do not have this problem since when the methods returns and the `BlkReq` is deallocated, it can be guaranteed that the virtio device has completed handling the request before.

I have no idea how to fix it yet.

qwandor:
I plan to fix this as part of some other improvements to buffer handling I'm working on. I'll send a PR soon.

wyfcyx:
This bug has been fixed in #40. Close it.

PR #40, opened 2022-12-20
Improve non-blocking API of VirtIOBlk and add documentation.

qwandor:
Pass request buffer into `read_block_nb` and `write_block_nb` rather than allocating it on the stack, so that it can be kept around until the request completes.

This fixes bug #7.


wyfcyx (approved):

wyfcyx:
LGTM, thank you!

PR #42, opened 2022-12-21
Make modules public rather than re-exporting all their contents

qwandor:
The crate namespace is getting rather full and difficult to follow. The Rustdoc is easier to read if we expose some of the modules directly.

I also added a device module for all the drivers for specific devices to live under, and some more documentation.


wangrunji0408 (approved):
The documentation is really helpful! Thanks!

PR #44, opened 2023-01-03
Try to connect to VM.

qwandor:
This should stop the virtio_net test from hanging forever waiting to recieve data, and fix the CI breakage.


chyyuu:
Thank you!

PR #46, opened 2023-01-05
Specify direction when allocating DMA regions

qwandor:
Only the legacy MMIO transport requires the three parts of the virtqueue to be contiguous. In other cases we can allocate the regions written by the driver separately from the region written by the device. This will allow systems with an IOMMU or equivalent to prevent the device from writing to regions which it shouldn't, such as the descriptor table and available ring.

This also reduces the size of the contiguous region allocated, which may reduce memory fragmentation.


chyyuu:
Very good work! @qwandor

PR #49, opened 2023-01-12
Make queue size a const generic parameter.

qwandor:
This fixes the issue where we were using the wrong size in AvailRing and UsedRing.


chyyuu:
Great work!

PR #50, opened 2023-01-12
Keep shadow copy of descriptor table.

qwandor:
This means we don't have to read back from the memory shared with the device, which is better in case a malicious device changes it.


chyyuu:
Great work!

PR #54, opened 2023-01-13
Support available buffer notification suppression, and send notifications for console and input

qwandor:
The `VirtIOConsole` and `VirtIOInput` drivers weren't sending available buffer notifications to the device like they should, so I've fixed that. I've also implemented support for the device to suppress available buffer notifications using the flags on the used ring.


chyyuu:
Great work!

PR #57, opened 2023-01-18
Allow HAL to validate MMIO ranges

qwandor:
The main goal of this change is to allow the `Hal` implementation to validate that the MMIO regions reported by a VirtIO device using the PCI transport is actually within the range expected by the OS. This prevents a malicious device from using bogus BARs to trick the driver into reading or writing arbitrary memory. This is important in a VM which doesn't trust its host, such as under pKVM.

`Hal::phys_to_virt` was previously being used for two different cases: for converting the physical addresses of DMA regions allocated by `Hal::dma_alloc` to virtual addresses, and for the MMIO addresses mentioned above. This PR removes the first use-case by having `Hal::dma_alloc` instead return both physical and virtual addresses for the allocated region. I've also added the virtual address as a parameter to `Hal::dma_dealloc` because it might be useful for some implementations. `Hal::phys_to_virt` is then renamed to `mmio_phys_to_virt` to reflect the only reason it is now used, and has an added size parameter so that the implementation can check that the whole region is in range.

I've also removed the `VirtAddr` type alias, because I think it makes more sense just to use a pointer everywhere that it was being used.

### DIFF
--- a/examples/aarch64/src/hal.rs
+++ b/examples/aarch64/src/hal.rs
@@ -4,7 +4,7 @@ use core::{
 };
 use lazy_static::lazy_static;
 use log::trace;
-use virtio_drivers::{BufferDirection, Hal, PhysAddr, VirtAddr, PAGE_SIZE};
+use virtio_drivers::{BufferDirection, Hal, PhysAddr, PAGE_SIZE};
 
 extern "C" {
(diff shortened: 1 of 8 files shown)


wangrunji0408 (approved):
Reasonable changes. LGTM!

PR #60, opened 2023-01-27
Mark parts of the `Hal` trait unsafe

qwandor:
The `Hal` trait should be marked as unsafe, as implementations need to return valid pointers from `dma_alloc` and `mmio_phys_to_virt` to avoid undefined behaviour in the drivers. Most of the functions in it (all except `dma_alloc`) should also be marked as unsafe, because the caller needs to pass valid pointers or addresses to avoid undefined behaviour in the HAL implementation. I've added comments explaining the requirements on each side.

I've also improved the safety comments in `queue.rs`, and changed various methods to take buffers by reference rather than pointer. This doesn't make them safe (except for `add_notify_wait_pop`, which I think actually is now), but reduces the safety requirements we need to document somewhat, as we can at least assume that a reference is non-null and is not being concurrently accessed.

### DIFF
--- a/examples/aarch64/src/hal.rs
+++ b/examples/aarch64/src/hal.rs
@@ -17,7 +17,7 @@ lazy_static! {
 
 pub struct HalImpl;
 
-impl Hal for HalImpl {
+unsafe impl Hal for HalImpl {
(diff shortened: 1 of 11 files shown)


wangrunji0408 (approved):
LGTM!

PR #61, opened 2023-02-16
Refactor virtio network device APIs to improve driver performance

equation314:
Main changes:
* Using the following APIs to receive/transmit packets:
  + receive: call `receive()` to pop a `RxBuffer` from the NIC queue, process its data, and then call `recycle_rx_buffer()` to add it back to the NIC queue.
  + transmit: call `new_tx_buffer()` to allocate a `TxBuffer`, fill it with data, and then call `send()` to transmit it.
* Pre-allocate `RxBuffer`s to fill the NIC queue on initialization.
* Add an example using [smoltcp](https://github.com/smoltcp-rs/smoltcp) that implements a simple echo server.

### DIFF
--- a/examples/riscv/Cargo.toml
+++ b/examples/riscv/Cargo.toml
@@ -6,10 +6,25 @@ edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+tcp = ["smoltcp"]
+default = []
+
 [dependencies]
 log = "0.4"
-riscv = "0.6"
+riscv = "0.10"
 opensbi-rt = { git = "https://github.com/rcore-os/opensbi-rt.git", rev = "abdfeb72" }
 fdt = "0.1.4"
 virtio-drivers = { path = "../.." }
 lazy_static = { version = "1.4", features = ["spin_no_std"] }
+
+[dependencies.smoltcp]
+version = "0.9.1"
+optional = true
+default-features = false
+features = [
+  "alloc", "log",   # no std
+  "medium-ethernet",
+  "proto-ipv4",
+  "socket-raw", "socket-icmp", "socket-udp", "socket-tcp",
+]
--- a/examples/riscv/Cargo.toml
+++ b/examples/riscv/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2018"
 
 [features]
 tcp = ["smoltcp"]
-default = []
+default = ["tcp"]
 
 [dependencies]
 log = "0.4"
--- a/src/device/mod.rs
+++ b/src/device/mod.rs
@@ -5,4 +5,5 @@ pub mod console;
 pub mod gpu;
 #[cfg(feature = "alloc")]
 pub mod input;
+#[cfg(feature = "alloc")]
 pub mod net;
--- a/src/device/net.rs
+++ b/src/device/net.rs
@@ -20,7 +20,7 @@ pub struct TxBuffer(Vec<u8>);
 
 /// A buffer used for receiving.
 pub struct RxBuffer {
-    buf: Vec<u8>,
+    buf: Vec<usize>, // for alignment
     packet_len: usize,
     idx: usize,
 }
@@ -51,7 +51,7 @@ impl RxBuffer {
     /// Allocates a new buffer with length `buf_len`.
     fn new(idx: usize, buf_len: usize) -> Self {
         Self {
-            buf: vec![0; buf_len],
+            buf: vec![0; buf_len / size_of::<usize>()],
             packet_len: 0,
             idx,
         }
@@ -69,13 +69,13 @@ impl RxBuffer {
 
     /// Returns all data in the buffer, including both the header and the packet.
     pub fn as_bytes(&self) -> &[u8] {
-        self.buf.as_slice()
+        self.buf.as_bytes()
     }
 
     /// Returns all data in the buffer with the mutable reference,
     /// including both the header and the packet.
     pub fn as_bytes_mut(&mut self) -> &mut [u8] {
-        self.buf.as_mut_slice()
+        self.buf.as_bytes_mut()
     }
 
     /// Returns the reference of the header.
@@ -85,12 +85,12 @@ impl RxBuffer {
 
     /// Returns the network packet as a slice.
     pub fn packet(&self) -> &[u8] {
-        &self.buf[NET_HDR_SIZE..NET_HDR_SIZE + self.packet_len]
+        &self.buf.as_bytes()[NET_HDR_SIZE..NET_HDR_SIZE + self.packet_len]
     }
 
     /// Returns the network packet as a mutable slice.
     pub fn packet_mut(&mut self) -> &mut [u8] {
-        &mut self.buf[NET_HDR_SIZE..NET_HDR_SIZE + self.packet_len]
+        &mut self.buf.as_bytes_mut()[NET_HDR_SIZE..NET_HDR_SIZE + self.packet_len]
     }
 }
 
@@ -106,7 +106,7 @@ pub struct VirtIONet<H: Hal, T: Transport, const QUEUE_SIZE: usize> {
     mac: EthernetAddress,
     recv_queue: VirtQueue<H, QUEUE_SIZE>,
     send_queue: VirtQueue<H, QUEUE_SIZE>,
-    rx_buffers: Vec<Option<RxBuffer>>,
+    rx_buffers: [Option<RxBuffer>; QUEUE_SIZE],
 }
 
 impl<H: Hal, T: Transport, const QUEUE_SIZE: usize> VirtIONet<H, T, QUEUE_SIZE> {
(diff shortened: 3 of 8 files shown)


examples/riscv/Cargo.toml
```diff
@@ -6,10 +6,25 @@ edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
+[features]
+tcp = ["smoltcp"]
+default = []
```
wangrunji0408, line 11: As an example crate, I'd like to have smoltcp enabled by default, so that users can see its usage with no more effort.

src/device/mod.rs
```diff
@@ -5,4 +5,5 @@ pub mod console;
 pub mod gpu;
 #[cfg(feature = "alloc")]
 pub mod input;
+#[cfg(feature = "alloc")]
 pub mod net;
```
wangrunji0408, line 9: Is it possible to keep a no-alloc subset of the net module?

wangrunji0408 (commented):
LGTM overall!

src/device/net.rs
```diff
@@ -4,13 +4,95 @@ use crate::hal::Hal;
...
+
+    /// Returns all data in the buffer with the mutable reference,
+    /// including both the header and the packet.
+    pub fn as_bytes_mut(&mut self) -> &mut [u8] {
+        self.buf.as_mut_slice()
+    }
+
+    /// Returns the reference of the header.
+    pub fn header(&self) -> &VirtioNetHdr {
+        unsafe { &*(self.buf.as_ptr() as *const VirtioNetHdr) }
```
qwandor, line 83: How do you know that `self.buf` satisfies the alignment requirements of `VirtioNetHdr`?

I think you can do this safely with `zerocopy::LayoutVerified::new_from_prefix(self.buf).unwrap().0.deref()`. It will panic if `self.buf` is not properly aligned though, so you still need to guarantee that somehow.

src/device/net.rs
```diff
@@ -41,11 +124,35 @@ impl<H: Hal, T: Transport> VirtIONet<H, T> {
...
         }
 
-        let recv_queue = VirtQueue::new(&mut transport, QUEUE_RECEIVE)?;
         let send_queue = VirtQueue::new(&mut transport, QUEUE_TRANSMIT)?;
+        let mut recv_queue = VirtQueue::new(&mut transport, QUEUE_RECEIVE)?;
+
+        let mut rx_buffers = Vec::with_capacity(QUEUE_SIZE);
+        for i in 0..QUEUE_SIZE {
+            let mut rx_buf = RxBuffer::new(i, buf_len);
+            let token = unsafe { recv_queue.add(&[], &mut [rx_buf.as_bytes_mut()])? };
```
qwandor, line 148: Can you add a comment here explaining how the safety requirements of `add` are guaranteed?

src/device/net.rs
```diff
@@ -77,32 +185,59 @@ impl<H: Hal, T: Transport> VirtIONet<H, T> {
...
-        // let header = unsafe { header.assume_init() };
-        Ok(len as usize - size_of::<Header>())
+    /// Receives a [`RxBuffer`] from network. If currently no data, returns an
+    /// error with type [`Error::NotReady`].
+    ///
+    /// It will try to pop a buffer that completed data reception in the
+    /// NIC queue.
+    pub fn receive(&mut self) -> Result<RxBuffer> {
+        if let Some(token) = self.recv_queue.peek_used() {
+            let mut rx_buf = self.rx_buffers[token as usize].take().unwrap();
```
qwandor, line 195: Rather than panicking if the buffer is `None`, which could happen if the device implementation is buggy, how about returning `Err(Error::WrongToken)`?

src/device/net.rs
```diff
@@ -77,32 +185,59 @@ impl<H: Hal, T: Transport> VirtIONet<H, T> {
...
+            assert_eq!(token as usize, rx_buf.idx);
+            let len = unsafe {
```
qwandor, line 197: A safety comment would be good here too.

src/device/net.rs
```diff
@@ -77,32 +185,59 @@ impl<H: Hal, T: Transport> VirtIONet<H, T> {
...
+                self.recv_queue
+                    .pop_used(token, &[], &mut [rx_buf.as_bytes_mut()])?
+            };
```
qwandor, line 200: Check that `len >= NET_HDR_SIZE` and return an error if not, as it indicates a buggy device.

src/device/net.rs
```diff
@@ -190,23 +325,43 @@ bitflags! {
...
     gso_type: GsoType,
     hdr_len: u16, // cannot rely on this
     gso_size: u16,
     csum_start: u16,
     csum_offset: u16,
+    // num_buffers: u16, // only available when the feature MRG_RXBUF is negotiated.
     // payload starts from here
 }
 
+impl Default for VirtioNetHdr {
```
qwandor, line 352: If you derive `Default` for `GsoType`, can you also just derive it for `VirtioNetHdr` rather than implementing it manually?

qwandor (commented):
Thanks, this looks like it should improve performance.

src/device/net.rs
chyyuu: @equation314 please concern and add comments about safety requirements on unsafe code.

src/device/net.rs
```diff
@@ -19,16 +101,17 @@ const QUEUE_SIZE: u16 = 2;
...
 /// A third command queue is used to control advanced filtering features.
-pub struct VirtIONet<H: Hal, T: Transport> {
+pub struct VirtIONet<H: Hal, T: Transport, const QUEUE_SIZE: usize> {
     transport: T,
     mac: EthernetAddress,
-    recv_queue: VirtQueue<H, { QUEUE_SIZE as usize }>,
-    send_queue: VirtQueue<H, { QUEUE_SIZE as usize }>,
+    recv_queue: VirtQueue<H, QUEUE_SIZE>,
+    send_queue: VirtQueue<H, QUEUE_SIZE>,
+    rx_buffers: Vec<Option<RxBuffer>>,
```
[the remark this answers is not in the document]
equation314: We must pin the buffers in memory, if use `[Option<RxBuffer>; QUEUE_SIZE]`, it may be moved if `VirtIONet` is moved.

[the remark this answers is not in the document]
qwandor: I don't think that's true, `RxBuffer` still contains a `Vec` so the array of pointers to the buffers will be moved but the buffers themselves won't be. Right now there are two levels of indirection, one should be enough.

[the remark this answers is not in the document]
equation314: I'm trying to put the `VirtioNetHdr` inside `RxBuffer` to resolve the alignment problem, so that will happen. How do you think to place the header?

src/device/net.rs
equation314: It never happens, when we add a buffer the the queue, it's index is equal to the token.

src/device/net.rs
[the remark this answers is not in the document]
qwandor: In that case maybe you can do something like:

```rust
pub struct RxBuffer {
  idx: usize,
  packet_length: usize,
  packet: Box<Packet>,
}

struct Packet {
  header: VirtioNetHdr,
  body: [u8],
}
```

Then `Packet` is unsized but should have the appropriate alignment, and the `Box` provides the indirection for it to be allocated on the heap. Allocation and deallocation might be tricky though, I haven't tried.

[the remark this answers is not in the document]
equation314: But in above we need to indicate the  size of `packet.body`.

src/device/net.rs
qwandor: I think what you want here and for the assertion below is actually not `token` but `token % QUEUE_SIZE`, otherwise this is going to crash after the first `QUEUE_SIZE` packets.

However, that aside, it's still possible for a buggy device to mark an invalid buffer as the next used one. Suppose that the device adds buffer 0 to the used ring, then immediately adds the same buffer to the used ring again before the driver marks it as available. It shouldn't do so, it is violating the VirtIO specification, but we shouldn't make our driver crash just because a device is buggy. It's better to return an error so we can try again or disable the device or something.

equation314: `token` is obtained from `VirtQueue::free_head`, which is obtained from `Descriptor::next`, which is never greater then or equal to `QUEUE_SIZE`.

qwandor: Ah you're right, sorry, the `% QUEUE_SIZE` isn't necessary. We probably should check that the device has sent a valid index though, and return an error rather than panicking if it is out of bounds.

src/device/mod.rs
equation314: May be another PR, their implementation can be really different.

examples/riscv/Cargo.toml
equation314: Enabling the tcp server will block CI.

chyyuu:
@equation314 LGTM ! 
@qwandor @wangrunji0408 How do you think?

PR #62, opened 2023-02-16
Allow Virtio Console driver to support multiple read requests

conradgrobler:
Since the `receive_token` was never reset, the virtio console device would only ever add one request. All future calls to `poll_retrieve` would do nothing.


qwandor (approved):
Thanks!

PR #65, opened 2023-03-23
Add protocols to support virtio socket device

aliciawyy:
With this change we can read the configuration space of a mounted virtio socket device by MMIO or PCI bus. The connection to the host still fails with the response VIRTIO_VSOCK_OP_RST.
    
I'll work on the virtqueue flows in subsequent changes.


src/endian.rs
```diff
@@ -0,0 +1,62 @@
+//! Endian types used in this library.
```
qwandor, line 1: What advantage do these have over the `zerocopy` types (https://docs.rs/zerocopy/latest/zerocopy/byteorder/index.html)?

aliciawyy: I changed to byteorder for explicit endian types. Could you help review this change?

src/device/socket/protocol.rs
```diff
@@ -0,0 +1,134 @@
...
+pub struct VirtioVsockEvent {
+    // ID from the virtio_vsock_event_id struct in the virtio spec
+    pub id: U32<LittleEndian>,
+}
+
+#[allow(non_camel_case_types)]
+#[derive(Copy, Clone, Eq, PartialEq, Debug)]
+#[repr(u16)]
+pub enum Op {
+    VIRTIO_VSOCK_OP_INVALID = 0,
```
qwandor, line 94: The `VIRTIO_VSOCK_OP_` prefix is unnecessary because enum variants are already namespaced in Rust, so you can just call them `Invalid` and so on.

src/device/socket/vsock.rs
```diff
@@ -0,0 +1,170 @@
...
+            info!("Device features: {:?}", features);
+            // negotiate these flags only
+            let supported_features = Feature::empty();
+            (features & supported_features).bits()
+        });
+
+        let config = transport.config_space::<VirtioVsockConfig>()?;
+        info!("config: {:?}", config);
+        // Safe because config is a valid pointer to the device configuration space.
+        let guest_cid = unsafe { volread!(config, guest_cid_low) as u64 };
```
qwandor, line 59: Shouldn't this read `guest_cid_high` too and combine them?

src/device/socket/vsock.rs
```diff
@@ -0,0 +1,170 @@
...
+                .pop_used(token, &[], &mut [&mut self.queue_buf_rx])?
+        };
+        let packet_rx = VirtioVsockPacket::read_from(&self.queue_buf_rx)?;
+        let result = match packet_rx.op()? {
+            Op::VIRTIO_VSOCK_OP_RESPONSE => Ok(()),
+            Op::VIRTIO_VSOCK_OP_RST => Err(SocketError::ConnectionFailed.into()),
+            Op::VIRTIO_VSOCK_OP_INVALID => Err(SocketError::InvalidOperation.into()),
+            _ => todo!(),
+        };
+        if result.is_err() {
```
qwandor, line 120: This seems like it should be the caller's responsibility to log if it wants to. If you want to log extra details here for debugging, how about just adding a `trace!("Received packet {:?}", packet_rx);` before the match? Then you don't need `result` either, the match can be the return value of the method.

src/device/socket/protocol.rs
```diff
@@ -0,0 +1,134 @@
...
+    pub hdr: VirtioVsockHdr,
+    pub data: &'a [u8],
+}
+
+impl<'a> VirtioVsockPacket<'a> {
+    pub fn read_from(buffer: &'a [u8]) -> Result<Self, SocketError> {
+        let hdr = buffer
+            .get(0..size_of::<VirtioVsockHdr>())
+            .ok_or(SocketError::BufferTooShort)?;
+        let hdr = VirtioVsockHdr::read_from(hdr).ok_or(SocketError::PacketParsingFailed)?;
```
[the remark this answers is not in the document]
aliciawyy: Thanks for the suggestion.

examples/aarch64/socket_test_host.py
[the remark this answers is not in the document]
aliciawyy: Actually the socket device can be read but the connection doesn't work yet. I can add a first setup and tune it later.

src/device/socket/protocol.rs
```diff
@@ -0,0 +1,134 @@
...
+use super::error::{self, SocketError};
+use crate::volatile::ReadOnly;
+use core::convert::TryInto;
+use core::{convert::TryFrom, mem::size_of};
+use zerocopy::{
+    byteorder::{LittleEndian, U16, U32, U64},
+    AsBytes, FromBytes,
+};
+
+pub const TYPE_STREAM_SOCKET: u16 = 1;
```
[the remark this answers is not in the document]
aliciawyy: Maybe still keep this as an enum for clarity?

examples/aarch64/vsock_server.cc
qwandor, line 1: Why is this in C++? Could it be in Rust, to keep things simple and consistent? The https://crates.io/crates/vsock crate could be helpful.

src/device/socket/protocol.rs
```diff
@@ -87,48 +98,62 @@ pub struct VirtioVsockEvent {
...
     /* Tell the peer our credit info */
-    VIRTIO_VSOCK_OP_CREDIT_UPDATE = 6,
+    CreditUpdate = 6,
     /* Request the peer to send the credit info to us */
-    VIRTIO_VSOCK_OP_CREDIT_REQUEST = 7,
+    CreditRequest = 7,
 }
 
-impl Into<U16<LittleEndian>> for Op {
+impl Into<U16<LittleEndian>> for VirtioVsockOp {
```
qwandor, line 121: Ditto, implement `From` rather than `Into`.

src/device/socket/vsock.rs
```diff
@@ -106,25 +108,19 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
         } else {
             return Err(SocketError::NoResponseReceived.into());
         };
+        // Safe because we are passing the same buffer as we passed to `VirtQueue::add`.
         let _len = unsafe {
             self.rx
                 .pop_used(token, &[], &mut [&mut self.queue_buf_rx])?
         };
         let packet_rx = VirtioVsockPacket::read_from(&self.queue_buf_rx)?;
```
qwandor, line 116: You don't need the `&` here, it is already a reference.

src/device/socket/protocol.rs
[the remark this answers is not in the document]
qwandor: Okay, if you prefer.

src/device/socket/protocol.rs
```diff
@@ -2,14 +2,28 @@
...
     byteorder::{LittleEndian, U16, U32, U64},
     AsBytes, FromBytes,
 };
 
-pub const TYPE_STREAM_SOCKET: u16 = 1;
+/// Currently only stream sockets are supported. type is 1 for stream socket types.
+/// Stream sockets provide in-order, guaranteed, connection-oriented delivery without message boundaries.
+#[derive(Copy, Clone, Debug)]
+pub enum SocketType {
+    Stream,
```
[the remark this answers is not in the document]
aliciawyy: Right. I forgot it when migrating the const to enum.

examples/aarch64/Makefile
```diff
@@ -14,6 +16,11 @@ ifeq ($(mode), release)
 	BUILD_ARGS += --release
 endif
 
+VSOCK_BUILD_ARGS = --target=x86_64-unknown-linux-gnu
```
qwandor, line 19: Is it necessary to specify this? It should just default to the host.

qwandor (approved):
Thanks! This looks good now, just one minor comment (and you'll need to rebase, sorry).

examples/aarch64/Makefile
qwandor: I see the issue, I've added a commit to fix it.

PR #67, opened 2023-03-30
Add #[allow(unused_unsafe)] to remove warnings.

aliciawyy:
This removes warnings like:
warning: unnecessary `unsafe` block


qwandor:
How about #68 instead?

PR #69, opened 2023-03-30
Fix bugs with device and queue reset

qwandor:
The VirtIO specificaiton says that for both device reset and MMIO transport queue reset the driver should read back 0 after writing 0, so block until it does. The PCI transport doesn't allow queue size to be set to 0, so there's no meaningful way we can reset it as far as I can see.


ptosi:
LGTM

PR #71, opened 2023-03-31
[vsock] Implement credit request/shutdown/send/recv

aliciawyy:


src/device/socket/vsock.rs
```diff
@@ -86,13 +97,40 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+        transport: &mut T,
+    ) -> Result {
+        let buffer_size = rx_buf.len() / QUEUE_SIZE;
+        if buffer_size < size_of::<VirtioVsockHdr>() {
+            return Err(SocketError::BufferTooShort.into());
+        }
+        for i in 0..QUEUE_SIZE {
+            let start = i * buffer_size;
+            let end = start + buffer_size;
+            // Safe because the buffer lives as long as the queue.
```
qwandor, line 118: I don't think you know that here. The reference being passed in only has a lifetime as long as the function call itself, but the queue most likely outlives that.

If you want to keep this function how it is, then it needs to be an unsafe function and the caller needs to guarantee that the buffer outlives the queue (and isn't used anywhere else as long as it is in the queue).

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
             op: VirtioVsockOp::Request.into(),
             ..Default::default()
         };
-        self.tx
-            .add_notify_wait_pop(&[header.as_bytes(), &[]], &mut [], &mut self.transport)?;
+        self.send_packet_to_tx_queue(&header, &[])?;
```
qwandor, line 142: Please add a comment here explaining why you're sending an empty packet, as it seems a bit surprising.

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+                self.connection_info = None;
+                Ok(())
+            }
+            _ => Err(SocketError::InvalidOperation.into()),
+        }
+    }
+
+    /// Waits until there is at least one element to pop in rx queue.
+    fn wait_one_in_rx_queue(&self) {
+        for _ in 0..10000000 {
```
qwandor, line 272: What is the significance of this number? It seems rather arbitrary.

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+            if self.rx.can_pop() {
+                break;
+            }
```
qwandor, line 275: If you really need to spin waiting for this, at least call `core::hint::spin_loop()` inside the loop.

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+                Ok(())
+            }
+            VirtioVsockOp::Rst => Err(SocketError::ConnectionFailed.into()),
+            _ => Err(SocketError::InvalidOperation.into()),
+        }
+    }
+
+    /// Sends the buffer to the destination.
+    pub fn send(&mut self, buffer: &[u8]) -> Result {
+        self.request_credit()?;
```
qwandor, line 198: Why do you need to call this every time before sending? I'm not super familiar with the details of vsock, but can't you check the free space you already know about first? And if there is no space available, I guess you shouldn't send the packet yet?

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+        let connection_info = self.connection_info()?;
+        let header = VirtioVsockHdr {
+            src_cid: self.guest_cid.into(),
+            dst_cid: connection_info.dst.cid.into(),
+            src_port: connection_info.src_port.into(),
+            dst_port: connection_info.dst.port.into(),
+            op: VirtioVsockOp::Rw.into(),
+            len: (buffer.len() as u32).into(),
+            buf_alloc: connection_info.buf_alloc.into(),
+            fwd_cnt: connection_info.fwd_cnt.into(),
```
qwandor, line 208: It appears that you are sending the host's buffer status back to it. Shouldn't we be telling it about our buffer status instead?

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+            src_cid: self.guest_cid.into(),
+            dst_cid: connection_info.dst.cid.into(),
+            src_port: connection_info.src_port.into(),
+            dst_port: connection_info.dst.port.into(),
+            op: VirtioVsockOp::CreditRequest.into(),
+            ..Default::default()
+        };
+        self.send_packet_to_tx_queue(&header, &[])?;
+        self.wait_one_in_rx_queue();
+        let received_header = self.pop_header_only_packet_from_rx_queue()?;
```
qwandor, line 176: Is it guaranteed that the next packet we receive after sending a credit request will be a credit update, or could it be interleaved with other packets?

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+            src_cid: self.guest_cid.into(),
+            dst_cid: connection_info.dst.cid.into(),
+            src_port: connection_info.src_port.into(),
+            dst_port: connection_info.dst.port.into(),
+            op: VirtioVsockOp::Rw.into(),
+            buf_alloc: connection_info.buf_alloc.into(),
+            fwd_cnt: connection_info.fwd_cnt.into(),
+            ..Default::default()
+        };
+        // TODO: We shouldn't need this send to trigger the fill of the rx queue.
```
qwandor, line 228: Why are you sending it then?

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+        self.send_packet_to_tx_queue(&header, &[])?;
+        self.wait_one_in_rx_queue();
+        let packet = self.pop_packet_from_rx_queue()?;
```
qwandor, line 231: How do you know that the next packet in the RX queue will be an RW, and not e.g. a CREDIT_UPDATE? The VirtIO spec says "It is also valid to send a VIRTIO_VSOCK_OP_CREDIT_UPDATE packet without previously receiving a VIRTIO_VSOCK_OP_CREDIT_REQUEST packet.", which suggests that we could receive one at any time.

src/device/socket/vsock.rs
```diff
@@ -86,13 +97,40 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+        rx_buf: &mut [u8],
+        transport: &mut T,
+    ) -> Result {
+        let buffer_size = rx_buf.len() / QUEUE_SIZE;
+        if buffer_size < size_of::<VirtioVsockHdr>() {
+            return Err(SocketError::BufferTooShort.into());
+        }
+        for i in 0..QUEUE_SIZE {
+            let start = i * buffer_size;
+            let end = start + buffer_size;
```
qwandor, line 117: Could this use `get_rx_buffer_range` too?

src/device/socket/vsock.rs
```diff
@@ -101,24 +139,185 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
-            _ => todo!(),
+        let _len = unsafe { self.rx.pop_used(token, &[], &mut [buffer])? };
+        let packet = VirtioVsockPacket::read_from(buffer)?;
+        debug!("Received packet {:?}. Op {:?}", packet, packet.hdr.op());
+        Ok(packet)
+    }
+
+    fn pop_header_only_packet_from_rx_queue(&mut self) -> Result<VirtioVsockHdr> {
+        let packet = self.pop_packet_from_rx_queue()?;
+        assert_eq!(packet.data.len(), 0);
```
qwandor, line 305: This should return an error rather than panicking, as it indicates a misbehaving device rather than a bug in the driver.

src/device/socket/vsock.rs
```diff
@@ -101,22 +139,124 @@ impl<'a, H: Hal, T: Transport> VirtIOSocket<'a, H, T> {
...
+        self.send_packet_to_tx_queue(&header, &[])?;
+        let received_header = self.get_header_only_packet_from_rx_queue()?;
+        match received_header.op()? {
+            VirtioVsockOp::Rst => {
+                info!("Disconnected from the peer");
+                self.connection_info = None;
+                Ok(())
+            }
+            VirtioVsockOp::Invalid => Err(SocketError::InvalidOperation.into()),
+            _ => todo!(),
```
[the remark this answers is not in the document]
qwandor: This was from an earlier version, ignore.

src/device/socket/vsock.rs
aliciawyy: It's possible actually. There're still some virtqueue controls not yet implemented, like I should process all the rx queue before sending out the request, and process the rx queue again when a packet arrives. That needs to be done for all these functions.

I would like to make the communication work first in this PR, then address these scenarios in a follow-up change. I added a TODO.

src/device/socket/vsock.rs
aliciawyy: It is arbitrary. I just want to set a time limit for the wait instead of endlessly spinning. How can we make this more clear?

examples/aarch64/Makefile
```diff
@@ -56,6 +56,19 @@ header: kernel
 clean:
 	cargo clean
 
+qemu-vsock: $(kernel_qemu_bin) $(img)
```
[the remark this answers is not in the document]
aliciawyy: This is for manual testing. If anyone wants to run the host listening socket in one shell and the guest in another, they can use this target. I also added a README.md in `/vsock_server/` to explain this.

examples/vsock_server/src/main.rs
```diff
@@ -1,13 +1,54 @@
...
+                    break;
+                }
+                Err(e) => {
+                    println!("{i} {e:?}");
+                    std::thread::sleep(Duration::from_millis(200))
+                }
+            }
+        }
+        println!("[Host] End.");
+        return;
```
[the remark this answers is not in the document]
aliciawyy: Actually I remove this return during the development to keep the server socket running for convenience. Just I'm not sure if it can pass the CI or not without the return.

examples/aarch64/Makefile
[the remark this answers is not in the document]
aliciawyy: I added a comment to explain this.

src/device/socket/vsock.rs
qwandor: Maybe the host is waiting for the guest to have capacity to receive, and sending an appropriate `VIRTIO_VSOCK_OP_CREDIT_UPDATE` would be enough?

aliciawyy: Right. That works. thanks.

src/device/socket/vsock.rs
aliciawyy: I adjusted the buf_alloc. About fwd_cnt, it looks like we need to cache the previous transmitted data length, maybe address that in a follow-up change together with the handling of the full queues?

qwandor (approved):

PR #72, opened 2023-04-05
[vsock] Minor adjustment (mostly about rx packet verification)

aliciawyy:


src/device/socket/vsock.rs
```diff
@@ -223,18 +218,15 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
 
     fn check_peer_buffer_is_sufficient(&mut self, buffer_len: usize) -> Result {
-        if usize::try_from(self.connection_info()?.peer_free())
-            .map_err(|_| SocketError::InvalidNumber)?
-            >= buffer_len
-        {
+        let buffer_len = buffer_len
+            .try_into()
+            .map_err(|_| SocketError::InvalidNumber)?;
+        if self.connection_info()?.peer_free() >= buffer_len {
```
qwandor, line 224: I think it makes more sense to keep the conversion the way it was. Converting `u32` to `usize` should never fail except on 16-bit platforms (which we don't support anyway), whereas converting `usize` to `u32` can fail. In fact there's no need for the `try_from`, a cast should be fine.

src/device/socket/vsock.rs
```diff
@@ -398,28 +382,34 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
-            assert_eq!(new_token, token);
-
+            if token != new_token {
+                return Err(SocketError::RecycledWrongBuffer.into());
+            }
             header_result
         }?;
 
+        // Notifies the RX queue after adding back the buffer.
+        if self.rx.should_notify() {
```
qwandor, line 402: I'd rather keep this outside of the loop for efficiency, so that it only gets called once rather than multiple times if we loop through several entries in the queue. Sending the notification will result in a context switch to the host VMM, so it's a relatively expensive operation.

src/device/socket/vsock.rs
```diff
@@ -398,28 +382,34 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
+            let header_result = read_header_and_body(buffer, body, packet_len);
 
             // Add the buffer back to the RX queue.
             let new_token = self.rx.add(&[], &mut [buffer])?;
             // If the RX buffer somehow gets assigned a different token, then our safety assumptions
             // are broken and we can't safely continue to do anything with the device.
-            assert_eq!(new_token, token);
-
+            if token != new_token {
+                return Err(SocketError::RecycledWrongBuffer.into());
```
qwandor, line 396: See the comment above. If we get the wrong token back here then our safety assumptions are broken and so anything more we do with the RX virtqueue buffers may be unsound, so we need to either panic or reset the device somehow, not just return an error and let the client keep using the device.

src/device/socket/vsock.rs
```diff
@@ -454,8 +444,15 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
 }
 
-fn read_header_and_body(buffer: &[u8], body: &mut [u8]) -> Result<VirtioVsockHdr> {
+fn read_header_and_body(
+    buffer: &[u8],
+    body: &mut [u8],
+    expected_packet_len: u32,
+) -> Result<VirtioVsockHdr> {
     let header = VirtioVsockHdr::read_from_prefix(buffer).ok_or(SocketError::BufferTooShort)?;
+    if expected_packet_len != header.packet_len()? {
```
qwandor, line 453: That doesn't look right. Shouldn't the VirtIO used buffer length match the total length of the data, including both the header and the body?

aliciawyy: Right. The packet_len() here is the total length including header and body.

aliciawyy:
It looks like this PR has lots of conflicts with the PR #73 you are working on. I'll close this one.

src/device/socket/vsock.rs
qwandor: Ah right sorry, I misread.

PR #73, opened 2023-04-05
Change vsock driver to poll model

qwandor:
Methods to connect, request credit and so on won't block until the other end responds to them, but rather will return immediately. The client should then call `poll_recv`, which will return one of several possible events either for data being received on the connection, or connection, disconnection. It can do this interleaved with whatever other operations it needs to do rather than blocking indefinitely on the vsock.

This also lays the groundwork for handling more than one connection at once, in which case events for different connections will be interleaved arbitrarily.

I've added some blocking helpers `wait_for_connect` and `wait_for_recv` to give something similar to the old behaviour, but these won't work with multiple connections.

### DIFF
--- a/src/device/socket/vsock.rs
+++ b/src/device/socket/vsock.rs
@@ -247,7 +247,7 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
     pub fn recv(&mut self, buffer: &mut [u8]) -> Result<usize> {
         let connection_info = self.connection_info()?;
 
-        // Tell the peer that we have space to recieve some data.
+        // Tell the peer that we have space to receive some data.
         let header = VirtioVsockHdr {
             op: VirtioVsockOp::CreditUpdate.into(),
             buf_alloc: (buffer.len() as u32).into(),
--- a/src/device/socket/vsock.rs
+++ b/src/device/socket/vsock.rs
@@ -397,9 +397,11 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
 
             match op {
                 VirtioVsockOp::Request => {
+                    header.check_data_is_empty()?;
                     // TODO: Send a Rst, or support listening.
                 }
                 VirtioVsockOp::Response => {
+                    header.check_data_is_empty()?;
                     return Ok(Some(VsockEvent {
                         source: connection_info.dst,
                         destination: VsockAddr {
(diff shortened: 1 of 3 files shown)


src/device/socket/vsock.rs
```diff
@@ -291,26 +364,21 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
         Ok(())
     }
 
-    fn poll_and_filter_packet_from_rx_queue<F>(
+    fn poll_and_filter_packet_from_rx_queue(
```
aliciawyy, line 367: We can remove `_and_filter` from the name since it doesn't filter anymore.

src/device/socket/vsock.rs
```diff
@@ -222,62 +271,86 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
-                >= buffer_len
-            {
-                Ok(())
-            } else {
-                Err(SocketError::InsufficientBufferSpaceInPeer.into())
+            // Request an update of the cached peer credit, if we haven't already done so, and tell
+            // the caller to try again later.
+            if !connection_info.credit_requested {
+                self.request_credit()?;
+                connection_info.credit_requested = true;
```
aliciawyy, line 286: So this means, if the guest calling send() fails with `SocketError::InsufficientBufferSpaceInPeer` error, then it needs to call `wait_for_recv()` first to get the credit updated, then it can call `send()` again, right?

aliciawyy (approved):
Thanks for the improvement!

src/device/socket/vsock.rs
qwandor: Yes. And it might need to handle receiving some other data in the meantime.

src/device/socket/vsock.rs
qwandor: Done.

src/device/socket/vsock.rs
aliciawyy: This approach may be implicit for an API. Although I understand that you want to avoid blocking the send() function, it may not be immediately obvious to users that a failed send() requires a call to recv() before retrying. Additionally, the usage of credit_requested may be unclear. Would it be possible to do a polling just after request_credit() ?

aliciawyy: I see. So the idea is to separate the interaction with the RX and TX queues into separate functions at the driver level, correct?

nit: Regarding `credit_requested`, I think renaming it to `has_pending_credit_request` could make its purpose clearer.

qwandor: I guess that's the implication, yes. I think it's probably still fine for `poll_recv` to send non-data packets to the TX virtqueue (e.g. responding to a credit request, shutdown or connection attempt), as that shouldn't block in most cases (though we do still need to figure out what to do in case the TX virtqueue is full, for some of these cases).

I've renamed `credit_requested` as you suggested.

PR #78, opened 2023-05-01
Fix some bugs

equation314:
1. No queue is selected when read `Transport::max_queue_size()`.
3. Device is not reset in `Transport::begin_init()`.

### DIFF
--- a/src/transport/pci.rs
+++ b/src/transport/pci.rs
@@ -231,10 +231,13 @@ impl Transport for PciTransport {
         }
     }
 
-    fn max_queue_size(&self) -> u32 {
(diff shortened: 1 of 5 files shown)


src/transport/pci.rs
```diff
@@ -231,10 +231,13 @@ impl Transport for PciTransport {
         }
     }
 
-    fn max_queue_size(&self) -> u32 {
+    fn max_queue_size(&self, queue: u16) -> u32 {
```
qwandor, line 234: This should now take `&mut self` rather than `&self`.

PR #79, opened 2023-05-03
Add x86_64 example using virtio PCI devices

equation314:
Add x86_64 example using virtio PCI devices:

- virtio-gpu-pci
- virtio-blk-pci
- virtio-net-pci

Currently the PCI ECAM base (MMCONFIG) is hardcoded to `0xB000_0000`.

### DIFF
--- a/examples/x86_64/linker.ld
+++ b/examples/x86_64/linker.ld
@@ -1,4 +1,4 @@
-OUTPUT_ARCH(riscv)
+OUTPUT_ARCH(i386:x86-64)
 ENTRY(_start)
 
 BASE_ADDRESS = 0x200000;
@@ -43,5 +43,5 @@ SECTIONS
     }
 
     . = ALIGN(4K);
-    PROVIDE(end = .);
+    PROVIDE(dma_region = .);
 }
--- a/examples/x86_64/linker.ld
+++ b/examples/x86_64/linker.ld
@@ -1,0 +1,47 @@
+OUTPUT_ARCH(riscv)
+ENTRY(_start)
+
+BASE_ADDRESS = 0x200000;
+
+SECTIONS
+{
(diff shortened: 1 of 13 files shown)


examples/x86_64/linker.ld
```diff
@@ -0,0 +1,47 @@
+OUTPUT_ARCH(riscv)
```
qwandor, line 1: Why riscv?

examples/x86_64/src/hal.rs
```diff
@@ -0,0 +1,50 @@
+use core::{
+    ptr::NonNull,
+    sync::atomic::{AtomicUsize, Ordering},
+};
+use lazy_static::lazy_static;
+use log::trace;
+use virtio_drivers::{BufferDirection, Hal, PhysAddr, PAGE_SIZE};
+
+extern "C" {
+    fn end();
```
qwandor, line 10: Why is this a function rather than a static variable? I guess either works as you just want to get the address, but it seems a bit weird.

examples/x86_64/src/logger.rs
```diff
@@ -0,0 +1,45 @@
...
+};
+
+struct Logger {
+    uart: SpinMutex<Option<SerialPort>>,
+}
+
+/// Initialises UART logger.
+pub fn init(max_level: LevelFilter) -> Result<(), SetLoggerError> {
+    // Safe because BASE_ADDRESS is the base of the MMIO region for a UART and is mapped as device
+    // memory.
```
qwandor, line 19: There's no MMIO or memory mapping involved here is there? I guess you mean that the I/O port belongs to a 16550 UART.

qwandor (approved):
I'm not very familiar with x86, but this seems reasonable.

PR #85, opened 2023-05-17
Use heap rather than DMA buffer for buffers which will be added to queues

qwandor:
Several device drivers were using `Dma::new` to allocate internal buffers which they then added to virtqueues. However, `VirtQueue::add` already takes care of sharing buffers with the device by calling `Hal::share`, which may copy the contents of the buffer to a DMA region if necessary, so the original buffer doesn't need to be allocated as a DMA buffer. In fact this may be a security issue in protected VMs, as the original buffer is shared with the host unnecessarily.

Instead, the internal buffers can just be allocated from the heap as usual.

This adds a dependency on `alloc` to the vsock, GPU and console drivers. If this is a problem then I could instead add a method to `Hal` to allocate a non-DMA buffer, but I think in practice most users of this crate will have `alloc` anyway.

### DIFF
--- a/src/device/socket/vsock.rs
+++ b/src/device/socket/vsock.rs
@@ -4,11 +4,12 @@
 use super::error::SocketError;
 use super::protocol::{VirtioVsockConfig, VirtioVsockHdr, VirtioVsockOp, VsockAddr};
 use crate::device::common::Feature;
-use crate::hal::{BufferDirection, Dma, Hal};
+use crate::hal::Hal;
 use crate::queue::VirtQueue;
 use crate::transport::Transport;
 use crate::volatile::volread;
-use crate::Result;
+use crate::{Result, PAGE_SIZE};
+use alloc::boxed::Box;
 use core::hint::spin_loop;
 use core::mem::size_of;
 use core::ptr::NonNull;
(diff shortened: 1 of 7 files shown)


src/device/socket/vsock.rs
```diff
@@ -108,7 +109,7 @@ pub struct VirtIOSocket<H: Hal, T: Transport> {
     /// The guest_cid field contains the guest’s context ID, which uniquely identifies
     /// the device for its lifetime. The upper 32 bits of the CID are reserved and zeroed.
     guest_cid: u64,
-    rx_buf_dma: Dma<H>,
+    rx_buf: NonNull<[u8; PAGE_SIZE]>,
```
equation314, line 112: Why use `NonNull` instead of `Box` here?

qwandor: `new` and `pop_packet_from_rx_queue` add subranges of it to the RX queue, so it is not safe to access it at the same time. Using a `Box` would allow safe code to access it, which would be unsound. Using a `NonNull` instead requires an `unsafe` block for any code which accesses it.

equation314 (approved):
LGTM

PR #91, opened 2023-05-25
Support multiple and incoming connections in vsock driver

qwandor:
This patch series starts by splitting out the connection management functionality of `VirtIOSocket` into a separate `SingleConnectionManager` which maintains the existing functionality. It then adds a more complex `VsockConnectionManager` which can handle multiple simultaneous connections, including incoming connections. `VsockConnectionManager` maintains per-connection ring buffers internally rather than copying directly from the virtqueue buffer to the client buffer as `SingleConnectionManager` does. This means an extra copying step, but allows it to implement credit properly and handle credit requests automatically. It also keeps track of which ports the guest is listening on, and automatically accepts connections on them while rejecting incoming connections on any other ports.

### DIFF
--- a/src/device/socket/multiconnectionmanager.rs
+++ b/src/device/socket/multiconnectionmanager.rs
@@ -151,7 +151,7 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
 
             if let VsockEventType::Received { length } = event.event_type {
                 // Copy to buffer
-                if !connection.buffer.write(body) {
+                if !connection.buffer.add(body) {
                     return Err(SocketError::OutputBufferTooShort(length).into());
                 }
             }
@@ -212,7 +212,7 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
         let (connection_index, connection) = get_connection(&mut self.connections, peer, src_port)?;
 
         // Copy from ring buffer
-        let bytes_read = connection.buffer.read(buffer);
+        let bytes_read = connection.buffer.drain(buffer);
 
         connection.info.done_forwarding(bytes_read);
 
@@ -325,7 +325,7 @@ impl RingBuffer {
     /// Adds the given bytes to the buffer if there is enough capacity for them all.
     ///
     /// Returns true if they were added, or false if they were not.
-    pub fn write(&mut self, bytes: &[u8]) -> bool {
+    pub fn add(&mut self, bytes: &[u8]) -> bool {
         if bytes.len() > self.available() {
             return false;
         }
@@ -346,7 +346,7 @@ impl RingBuffer {
 
     /// Reads and removes as many bytes as possible from the buffer, up to the length of the given
     /// buffer.
-    pub fn read(&mut self, out: &mut [u8]) -> usize {
+    pub fn drain(&mut self, out: &mut [u8]) -> usize {
         let bytes_read = min(self.used, out.len());
 
         // The number of bytes to copy out between `start` and the end of the buffer.
--- a/src/device/socket/multiconnectionmanager.rs
+++ b/src/device/socket/multiconnectionmanager.rs
@@ -252,3 +252,240 @@ impl RingBuffer {
         bytes_read
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        device::socket::{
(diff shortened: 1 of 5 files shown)


src/device/socket/multiconnectionmanager.rs
```diff
@@ -0,0 +1,763 @@
...
+            .copy_from_slice(&bytes[0..write_before_wraparound]);
+        self.buffer[0..write_after_wraparound].copy_from_slice(&bytes[write_before_wraparound..]);
+        self.used += bytes.len();
+
+        true
+    }
+
+    /// Reads and removes as many bytes as possible from the buffer, up to the length of the given
+    /// buffer.
+    pub fn read(&mut self, out: &mut [u8]) -> usize {
```
aliciawyy, line 350: Since the buffer will be dropped after reading, what about renaming this to `pop` and the `write` function `push`?

src/device/socket/multiconnectionmanager.rs
```diff
@@ -0,0 +1,763 @@
...
+        if bytes.len() > self.available() {
+            return false;
+        }
+
+        let end = (self.start + self.used) % self.buffer.len();
+        let write_before_wraparound = min(bytes.len(), self.buffer.len() - end);
+        let write_after_wraparound = bytes
+            .len()
+            .checked_sub(write_before_wraparound)
+            .unwrap_or_default();
```
aliciawyy, line 339: Maybe this part 
```
let write_after_wraparound = bytes
            .len()
            .checked_sub(write_before_wraparound)
            .unwrap_or_default();
self.buffer[0..write_after_wraparound].copy_from_slice(&bytes[write_before_wraparound..]);
```
can be refactored as following to be more clear?
```
bytes.get(bytes_split_index..)
    .map(|remaining| self.buffer[..remaining.len()].copy_from_slice(remaining));
```

src/device/socket/multiconnectionmanager.rs
```diff
@@ -0,0 +1,763 @@
...
+            }
+            VsockEventType::CreditUpdate => {}
+        }
+
+        Ok(Some(event))
+    }
+
+    /// Reads data received from the given connection.
+    pub fn recv(&mut self, peer: VsockAddr, src_port: u32, buffer: &mut [u8]) -> Result<usize> {
+        let (connection_index, connection) = self
```
aliciawyy, line 223: This part repeats 4 times in the code, what about extracting a function for it, something like `get_connection_info(peer, port)`?

src/device/socket/vsock.rs
```diff
@@ -348,173 +419,83 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
...
-            }
+    /// # Safety
+    ///
+    /// The buffer must not currently be in the RX queue, and no other references to it must exist
+    /// between when this method is called and when it is popped from the queue.
+    unsafe fn add_buffer_to_rx_queue(&mut self, index: u16) -> Result {
+        // Safe because the buffer lives as long as the queue, and the caller guarantees that it's
+        // not currently in the queue or referred to anywhere else until it is popped.
+        unsafe {
+            let buffer = self.rx_queue_buffers[usize::from(index)].as_mut();
```
aliciawyy, line 432: Use `.get` to get the mut buffer just to be sure?

aliciawyy (approved):
LGTM overall. It's a good idea to move the ConnectionInfo out of the device driver itself.

aliciawyy (approved):
LGTM overall. It's a good idea to move the ConnectionInfo out of the device driver itself.

src/device/socket/multiconnectionmanager.rs
aliciawyy: It's true this is more of a queue, I think `add` is okay but I don't see `drain` so much used for queues. What about `add` and `remove` like the Queue API in java?

src/device/socket/multiconnectionmanager.rs
```diff
@@ -0,0 +1,763 @@
...
+    ///
+    /// Returns true if they were added, or false if they were not.
+    pub fn write(&mut self, bytes: &[u8]) -> bool {
+        if bytes.len() > self.available() {
+            return false;
+        }
+
+        let end = (self.start + self.used) % self.buffer.len();
+        let write_before_wraparound = min(bytes.len(), self.buffer.len() - end);
+        let write_after_wraparound = bytes
```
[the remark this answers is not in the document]
aliciawyy: LGTM

PR #93, opened 2023-06-13
New virtio blk driver

Arsering:


src/lib.rs
```diff
@@ -51,7 +51,7 @@ extern crate alloc;
 
 pub mod device;
 mod hal;
-mod queue;
+mod virtio_queue;
```
qwandor, line 54: Why the rename? We already know the `queue` module is referring to a virtio queue, as it's part of the `virtio-drivers` crate.

src/lib.rs
```diff
@@ -145,3 +145,5 @@ fn pages(size: usize) -> usize {
 fn nonnull_slice_from_raw_parts<T>(data: NonNull<T>, len: usize) -> NonNull<[T]> {
     NonNull::new(ptr::slice_from_raw_parts_mut(data.as_ptr(), len)).unwrap()
 }
+
+
```
qwandor, line 149: Nit: remove the extra blank line.

qwandor (commented):
Thanks for the PR! It looks like you've forgotten to include a bunch of new files though.

PR #100, opened 2023-07-11
Add support for flushing block devices and getting their IDs

qwandor:
This is part of an attempt at splitting #96 into more manageable chunks to review.

These features are fairly straightforward, and don't change anything about memory management or queue structure.


equation314 (approved):
LGTM

PR #102, opened 2023-07-13
Add support for indirect descriptors in VirtQueue with alloc feature

qwandor:
This is based on some of the work in #96, but fixes some issues there with how allocation and DMA memory is handled, adds tests for the indirect virtqueue itself, and also adds support to the fake device support so that device tests will work with indirect queues.

Devices can choose when constructing a `VirtQueue` whether it should support indirect descriptors or not, based on their preference and the device features. For now only the `VirtIOBlk` driver uses indirect descriptors (if the underlying device supports it). If a `VirtQueue` is constructed with indirect descriptor support then it will use them whenever more than one buffer is added to the queue. There's no point using indirect descriptors for a single buffer so we don't bother.

Space for the indirect descriptor lists is allocated (and shared with the device via the HAL) dynamically. This means that it depends on the `alloc` feature, which is enabled by default. If the `alloc` feature is disabled, then indirect descriptors will never be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ VirtIO guest drivers in Rust. For **no_std** environment.
 
 | Feature flag                 | Supported |                                         |
 | ---------------------------- | --------- | --------------------------------------- |
-| `VIRTIO_F_INDIRECT_DESC`     | ❌        | Indirect descriptors                    |
+| `VIRTIO_F_INDIRECT_DESC`     | ✅        | Indirect descriptors                    |
 | `VIRTIO_F_EVENT_IDX`         | ❌        | `avail_event` and `used_event` fields   |
 | `VIRTIO_F_VERSION_1`         | TODO      | VirtIO version 1 compliance             |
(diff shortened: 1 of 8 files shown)


equation314 (approved):
LGTM

equation314 (commented):
Maybe we should add a new feature "indirect-desc"?

qwandor:
> Maybe we should add a new feature "indirect-desc"?

You mean a cargo feature? Is there any case where you have an allocator available, the VirtIO device implementation supports indirect descriptors, but you wouldn't want to use them? I can't think of one, but if so we could add a flag to the device driver constructor so it could be decided at runtime.

equation314:
The conditional compilation attribute `#[cfg(feature = "alloc")]` is confused, I think use `#[cfg(feature = "indirect-desc")]` is more better.

equation314:
Okay, I mean we can enable it by default and make it dependent on `alloc`

PR #103, opened 2023-07-18
Avoid adding empty buffers to queue

qwandor:
The vsock driver was doing this, which caused problems with the HAL implementation trying to share an empty region of memory.

### DIFF
--- a/src/device/socket/vsock.rs
+++ b/src/device/socket/vsock.rs
@@ -411,11 +411,16 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
     }
 
     fn send_packet_to_tx_queue(&mut self, header: &VirtioVsockHdr, buffer: &[u8]) -> Result {
-        let _len = self.tx.add_notify_wait_pop(
-            &[header.as_bytes(), buffer],
-            &mut [],
-            &mut self.transport,
-        )?;
+        let _len = if buffer.is_empty() {
+            self.tx
+                .add_notify_wait_pop(&[header.as_bytes()], &mut [], &mut self.transport)?
+        } else {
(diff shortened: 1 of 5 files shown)


qwandor:
@aliciawyy, PTAL.

aliciawyy:
net driver can also send empty buffer in theory.

https://github.com/rcore-os/virtio-drivers/blob/master/src/device/net.rs#L247

src/device/socket/vsock.rs
```diff
@@ -411,11 +411,16 @@ impl<H: Hal, T: Transport> VirtIOSocket<H, T> {
     }
 
     fn send_packet_to_tx_queue(&mut self, header: &VirtioVsockHdr, buffer: &[u8]) -> Result {
-        let _len = self.tx.add_notify_wait_pop(
-            &[header.as_bytes(), buffer],
-            &mut [],
-            &mut self.transport,
-        )?;
+        let _len = if buffer.is_empty() {
```
aliciawyy, line 414: Why do we need `let _len =` instead of using `if ... else` directly?

aliciawyy (approved):

src/device/socket/vsock.rs
qwandor: It doesn't do anything, it just makes it explicit that we are ignoring the returned length. We had this before and it seemed worthwhile to keep.

qwandor:
> net driver can also send empty buffer in theory.
> 
> https://github.com/rcore-os/virtio-drivers/blob/master/src/device/net.rs#L247

Added a special case for this too, though it seems like a weird thing for someone to want to do.

PR #109, opened 2023-08-08
Remove SingleConnectionManager

qwandor:
There's no good reason to use it over `VsockConnectionManager`, and it doesn't implement receive buffer credit properly.


src/device/socket/mod.rs
```diff
@@ -7,20 +7,16 @@
...
 //! See [`VsockConnectionManager`] for a usage example.
 
-mod error;
 #[cfg(feature = "alloc")]
-mod multiconnectionmanager;
+mod connectionmanager;
+mod error;
 mod protocol;
 #[cfg(feature = "alloc")]
-mod singleconnectionmanager;
```
aliciawyy, line 15: The comment above needs to be updated as well

aliciawyy (approved):
thanks

src/device/socket/mod.rs
qwandor: Thanks for spotting that! Fixed.

PR #110, opened 2023-08-24
[vsock] Add method to send credit update manually

aliciawyy:
This PR addresses a bug in the vsock connection manager. Prior to the PR, the guest doesn't update the host after it processes the local buffer. As a result, when the receiving message length reaches the buffer capacity, the host could mistakenly stop sending the message, unaware that the guest still had available buffer.

This fix ensures that the credit is correctly updated after processing the local buffer, allowing for continuous message transmission between the host and the guest.

### DIFF
--- a/src/device/socket/connectionmanager.rs
+++ b/src/device/socket/connectionmanager.rs
@@ -525,7 +525,7 @@ mod tests {
                 .unwrap()
                 .write_to_queue::<QUEUE_SIZE>(RX_QUEUE_IDX, &response);
 
-            // Expects a credit update.
+            // Expect a credit update.
             State::wait_until_queue_notified(&state, TX_QUEUE_IDX);
             assert_eq!(
                 VirtioVsockHdr::read_from(
@@ -549,7 +549,7 @@ mod tests {
                     fwd_cnt: (hello_from_host.len() as u32).into(),
                 }
             );
-            // Expects a shutdown.
+            // Expect a shutdown.
             State::wait_until_queue_notified(&state, TX_QUEUE_IDX);
             assert_eq!(
                 VirtioVsockHdr::read_from(
--- a/src/device/socket/connectionmanager.rs
+++ b/src/device/socket/connectionmanager.rs
@@ -525,7 +525,7 @@ mod tests {
                 .unwrap()
                 .write_to_queue::<QUEUE_SIZE>(RX_QUEUE_IDX, &response);
 
-            // Expect a credit update.
+            // Expects a credit update.
             State::wait_until_queue_notified(&state, TX_QUEUE_IDX);
             assert_eq!(
                 VirtioVsockHdr::read_from(
--- a/src/device/socket/connectionmanager.rs
+++ b/src/device/socket/connectionmanager.rs
@@ -525,7 +525,31 @@ mod tests {
                 .unwrap()
                 .write_to_queue::<QUEUE_SIZE>(RX_QUEUE_IDX, &response);
 
-            // Expect a shutdown.
+            // Expects a credit update.
+            State::wait_until_queue_notified(&state, TX_QUEUE_IDX);
+            assert_eq!(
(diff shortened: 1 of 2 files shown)


aliciawyy:
It looks like there's some connection issues with clippy in CI. Can someone take a look at this PR?

src/device/socket/connectionmanager.rs
```diff
@@ -213,8 +213,10 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
 
         // Copy from ring buffer
         let bytes_read = connection.buffer.drain(buffer);
-
-        connection.info.done_forwarding(bytes_read);
+        if bytes_read > 0 {
+            connection.info.done_forwarding(bytes_read);
+            self.driver.credit_update(&connection.info)?;
```
qwandor, line 218: I don't think this is a good idea. It what was previously a non-blocking call (except in the case that the connection is in the process of shutdown) into a blocking call. This will significantly reduce performance, especially if the client is calling `recv` many times in a row with small buffers. It's also sending lots of credit updates that the host didn't ask for, which while allowed by the specification is unneccessary.

Credit is already updated for the connection whenever we send anything else on it (e.g. sending some data), which is likely to be enough a fair bit of the time in practice. If the host is blocked waiting for credit then it should send a credit request, which `VsockConnectionManager::poll` will automatically respond to. So all the client needs to do is ensure they are regularly calling `poll`, which is anyway necessary to avoid other deadlocks.

aliciawyy: > Credit is already updated for the connection whenever we send anything else on it (e.g. sending some data), which is likely to be enough a fair bit of the time in practice.

That's true. But it is still blocked when the guest just wants to receive a message longer than 1024 bytes and it sends nothing to the host till it receives the whole message.

> If the host is blocked waiting for credit then it should send a credit request ...

It looks like in practice the host doesn't always send credit request when it gets blocked. You can reproduce the failure without this fix by increasing the message length in this [test](https://cs.android.com/android/platform/superproject/+/master:packages/modules/Virtualization/rialto/tests/test.rs;l=188;drc=748b032fb17d902378d523505413885943cc93dc).

aliciawyy: What about I just add a new API to update the credit to allow users to call it by themselves?

qwandor: Hmm, that's weird. Are you seeing this behaviour both with QEMU and crosvm?

Adding an API to send an unrequested credit update is alright I guess. I still think you shouldn't call it for every receive, as that will have a significant impact on performance. Perhaps you could just call it once you've emptied the receive buffer? To achieve that you could expose a new method on `VsockConnectionManager` to return the number of bytes currently in the receive buffer (perhaps call it `available`), and if after receiving a non-zero number of bytes `available` returns 0, then you know that the buffer has just become empty. Further optimisations are probably possible.

aliciawyy: The problem is with crosvm.

src/device/socket/connectionmanager.rs
```diff
@@ -226,6 +226,18 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
         Ok(bytes_read)
     }
 
+    /// Returns whether the receive buffer is empty.
+    pub fn is_recv_buffer_empty(&mut self, peer: VsockAddr, src_port: u32) -> Result<bool> {
```
qwandor, line 230: I think an `available` method which returns the number of bytes available to read (as I was suggesting before) would be better, as it's useful for other things as well.

aliciawyy: Done

qwandor (approved):

aliciawyy:
Hi @qwandor can we release a new patch with these APIs?

qwandor:
> Hi @qwandor can we release a new patch with these APIs?

Done.

PR #114, opened 2023-10-11
Change the function to retrieve used buffer instead of available buffer

aliciawyy:
The status of the used receive buffer is more helpful for users when configuring credit updates.

### DIFF
--- a/src/device/socket/connectionmanager.rs
+++ b/src/device/socket/connectionmanager.rs
@@ -226,13 +226,13 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
         Ok(bytes_read)
     }
 
-    /// Returns the number of used bytes in the receive buffer.
+    /// Returns the number of bytes in the receive buffer available to be read by `recv`.
     ///
-    /// When the used bytes is 0, it indicates that the receive buffer is empty and does not
+    /// When the available bytes is 0, it indicates that the receive buffer is empty and does not
     /// contain any data.
-    pub fn recv_buffer_used_bytes(&mut self, peer: VsockAddr, src_port: u32) -> Result<usize> {
+    pub fn recv_buffer_available_bytes(&mut self, peer: VsockAddr, src_port: u32) -> Result<usize> {
         let (_, connection) = get_connection(&mut self.connections, peer, src_port)?;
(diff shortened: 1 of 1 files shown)


src/device/socket/connectionmanager.rs
```diff
@@ -226,10 +226,13 @@ impl<H: Hal, T: Transport> VsockConnectionManager<H, T> {
         Ok(bytes_read)
     }
 
-    /// Returns the number of bytes currently available in the recv buffer.
-    pub fn recv_buffer_available_bytes(&mut self, peer: VsockAddr, src_port: u32) -> Result<usize> {
+    /// Returns the number of used bytes in the receive buffer.
+    ///
+    /// When the used bytes is 0, it indicates that the receive buffer is empty and does not
+    /// contain any data.
+    pub fn recv_buffer_used_bytes(&mut self, peer: VsockAddr, src_port: u32) -> Result<usize> {
```
qwandor, line 233: The behaviour is now good, but I think we should call it available, because from the point of view of the caller it is the number of bytes which are available to read. I'll upload another commit to fix this.

qwandor (approved):

src/device/socket/connectionmanager.rs
aliciawyy: I feel it confusing to call it available. Especially the same word is used in the same function to indicate the free space in buffer .

qwandor: Fair point about the method on `RingBuffer`, we could rename that to `free` or something if you like. But I really think that 'available' is the best word for the number of bytes which are ready to be read by the caller. The caller shouldn't usually need to care about the details of how the buffering works, all it needs to know is how much data is available to read.

aliciawyy: > The caller shouldn't usually need to care about the details of how the buffering works, all it needs to know is how much data is available to read.

Agreed. I just don't like using "available" twice in the same code to convey two contradictory meanings for the buffer. It's easier to understand if we adopt consistent naming, or include `available_to_read` in the name.

qwandor: Okay, uploaded #115, PTAL.

PR #115, opened 2023-10-11
Rename RingBuffer#available to free.

qwandor:
This avoids using the same name as recv_buffer_available_bytes.


aliciawyy:
Can we publish a minor patch with the API update?

qwandor:
> Can we publish a minor patch with the API update?

Done.

aliciawyy:
Thank you

PR #117, opened 2023-11-06
! upgrade riscv demo rust-toolchain's version to nightly

lispking:
rt

### DIFF
--- a/examples/riscv/rust-toolchain
+++ b/examples/riscv/rust-toolchain
@@ -1,1 +1,1 @@
-nightly-2022-11-03
+nightly


examples/riscv/rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2022-11-03
\ No newline at end of file
+nightly
```
qwandor, line 1: Is there any chance we can get this onto a stable toolchain?

lispking: **The stable branch will report an error because linked_list_allocator unsupported stable.**

```shell
Compiling linked_list_allocator v0.9.1
error[E0554]: `#![feature]` may not be used on the stable release channel
1 | #![cfg_attr(feature = "const_mut_refs", feature(const_mut_refs))]
```


Check if the dependency count is older for packages that rely on opensbi-rt

Another interesting thing is that this library doesn't seem to be officially planned for maintenance anymore?

https://os.phil-opp.com/kernel-heap/

`cargo tree`

```shell
├── opensbi-rt v0.1.0 (https://github.com/rcore-os/opensbi-rt.git?rev=abdfeb72#abdfeb72)
│   ├── **linked_list_allocator v0.9.1**
│   │   └── spinning_top v0.2.5
│   │       └── lock_api v0.4.11
│   │           └── scopeguard v1.2.0
│   │           [build-dependencies]
│   │           └── autocfg v1.1.0
│   ├── log v0.4.20
│   └── spin v0.9.8
│       └── lock_api v0.4.11 (*)
```

qwandor: You could look into using `buddy_system_allocator`, as in the aarch64 example. However, that doesn't need to block merging this PR, I'm happy to merge once the presubmit checks and fixed.

PR #111, opened 2023-09-25
Introduce VirtIONetRaw to allow custom NIC buffer management and used buffer notification suppression

hky1999:
Some tiny modifications need to be completed within ArceOS. @equation314

PR #119, opened 2023-11-28
ci: turn on tcp feature in examples

lwshang:
When I was testing the x86_64 example with `tcp` feature, I got a compilation error at the line of `dev.transmit(tx_buf).unwrap();`.

The error was introduced in a recent PR: https://github.com/rcore-os/virtio-drivers/pull/111.
It was overlooked by CI because the default value `tcp=off` was in effect.

I found the original rationale for disabling `tcp` in CI from this commit message: https://github.com/rcore-os/virtio-drivers/pull/86/commits/5b0786b09ec94840ad181818be72ba62d217dee0
>Disable tcp feature in x86_64 example by default.
> It waits forever, which is no good in CI.

So I modify the `test_echo_server` function to break out the dead loop.

I also applied the same fix to the riscv example.

### DIFF
--- a/examples/aarch64/rust-toolchain.toml
+++ b/examples/aarch64/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "nightly"
+channel = "stable"
 components = ["rustfmt", "llvm-tools"]
 targets = ["aarch64-unknown-none"]
 profile = "minimal"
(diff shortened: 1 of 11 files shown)


lwshang:
GitHub CI have `rustup` by default, it can pickup the correct toolchain from `rust-toolchain.toml` files automatically.

So there is no need to setup toolchain in `main.yaml` anymore. The toml files will be the only source of truth.

examples/aarch64/rust-toolchain.toml
```diff
@@ -0,0 +1,5 @@
+[toolchain]
+channel = "nightly"
```
qwandor, line 2: Can we keep this on stable? It was working with stable before, I'd like to keep it that way.

qwandor (approved):

qwandor:
Thanks!

PR #126, opened 2024-04-22
Add Debug trait implementation for ReadOnly, WriteOnly, and Volatile

fslongjin:


src/volatile.rs
```diff
@@ -104,5 +122,7 @@ macro_rules! volwrite {
     };
 }
 
+use core::fmt::{Debug, Formatter};
```
qwandor, line 125: Put these imports at the top of the file.

src/volatile.rs
```diff
@@ -3,6 +3,12 @@
 #[repr(transparent)]
 pub struct ReadOnly<T: Copy>(T);
 
+impl<T: Debug + Copy> Debug for ReadOnly<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
```
qwandor, line 7: No need for the explicit lifetime `<'_>` here and in the other implementation below.

src/volatile.rs
```diff
@@ -15,6 +21,12 @@ impl<T: Copy> ReadOnly<T> {
 #[repr(transparent)]
 pub struct WriteOnly<T: Copy>(T);
 
+impl<T: Debug + Copy> Debug for WriteOnly<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("WriteOnly").field(&self.0).finish()
```
qwandor, line 26: `WriteOnly` is used to wrap MMIO registers which can only be written, so it doesn't make sense to read a value from it.

src/volatile.rs
```diff
@@ -3,6 +3,12 @@
 #[repr(transparent)]
 pub struct ReadOnly<T: Copy>(T);
 
+impl<T: Debug + Copy> Debug for ReadOnly<T> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+        f.debug_tuple("ReadOnly").field(&self.0).finish()
```
qwandor, line 8: `ReadOnly` is intended to wrap MMIO registers which may have side-effects when reading, so it's probably not a good idea to implement `Debug`, as adding debug output may change behaviour. Also any reading should be done with `read_volatile`.

qwandor (commented):
How would these `Debug` implementations be used? These types are used to wrap MMIO registers, which can't just be read directly.

src/volatile.rs
fslongjin: Indeed, you are right, I didn't consider the issue of side effects.

PR #128, opened 2024-04-22
Use atomic reads and writes for available and used ring.

### LINKED ISSUE #125: The examples/riscv has a stuck bug
Running examples/riscv will get stuck.

Like below:

屏幕截图 2024-04-12 161449

qwandor:
This is necessary to correctly synchronise with the host. The previous fences were not correct.

Fixes #125.


fslongjin:
It run wells in both `examples/riscv` and my test case.

PR #136, opened 2024-05-30
 Allow vsock stream shutdown hints to be specified

qwandor:
Linux seems to require both send and receive shutdown hints before it will acknowledge the shutdown, so set both by default.


thedataking (approved):
LGTM 👍

PR #137, opened 2024-06-03
Added sound driver based on virtio v1.2.

muxinyu1:
- Only the output stream has been implemented.
- Two audio playback modes have been implemented: synchronous blocking and asynchronous non-blocking.
- Tested successfully on Ubuntu 22.04 LTS.


src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    jacks: ReadOnly<u32>,
+    streams: ReadOnly<u32>,
+    chmaps: ReadOnly<u32>,
+}
+
+/// In virtIO v1.2, this enum should be called "Code".
+///
+/// To avoid ambiguity in its meaning, I use the term "CommandCode" here.
+#[repr(u32)]
+#[derive(FromPrimitive, IntoPrimitive)]
```
qwandor, line 1090: Derive `Copy` and `Clone` too, and maybe `Debug`, `Eq` and `PartialEq` in case they come in handy later for debugging or testing.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    VirtioSndSBadMsg,
+    /// requested operation or parameters are not supported
+    VirtioSndSNotSupp,
+    ///  an I/O error occurred
+    VirtioSndSIoErr,
+}
+
+/// Enum representing the types of item information requests.
+#[derive(Clone, Copy)]
+pub enum ItemInfomationRequestType {
```
qwandor, line 1129: This looks like it should also be `#[repr(u32)]`. Also, you've made a typo in the name, it should be `ItemInformationRequestType`.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    /// a control message is malformed or contains invalid parameters
+    VirtioSndSBadMsg,
+    /// requested operation or parameters are not supported
+    VirtioSndSNotSupp,
+    ///  an I/O error occurred
+    VirtioSndSIoErr,
+}
+
+/// Enum representing the types of item information requests.
+#[derive(Clone, Copy)]
```
qwandor, line 1128: Derive `Debug`, `Eq` and `PartialEq` too, as users of the library may find them useful for debugging and testing.

src/device/sound.rs
qwandor: Actually, how is this different to `CommandCode`? It seems like a subset of it.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    /// An underflow for the output stream or an overflow for the inputstream has occurred.
+    PcmXrun,
+    /// An external device has been connected to the jack.
+    JackConnected = 0x1000,
+    /// An external device has been disconnected from the jack.
+    JackDisconnected,
+}
+
+/// Notification from sound device.
+pub struct Notification {
```
qwandor, line 1207: Derive `Clone`, `Debug`, `Eq` and `PartialEq` as users of the library may find them useful.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    VirtioSndPcmFmtIec958Subframe, /* 32 / 32 bits */
+}
+
+impl From<PcmSampleFormat> for u64 {
+    fn from(value: PcmSampleFormat) -> Self {
+        value as _
+    }
+}
+
+enum PcmFrameRate {
```
qwandor, line 1376: This doesn't seem to be used anywhere, remove it.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+}
+
+impl Display for VirtIOSndPcmInfo {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let mut features = String::new();
+        bitflags::parser::to_writer(&PcmFeatures::from(self.features), &mut features).unwrap();
+        let mut rates = String::new();
+        let _ = bitflags::parser::to_writer(&PcmRate::from(self.rate), &mut rates).unwrap();
+        let mut formats = String::new();
+        let _ = bitflags::parser::to_writer(&PcmFormats::from(self.formats), &mut formats).unwrap();
```
qwandor, line 1423: Derive `Debug` for the various bitflags types and use that rather than allocating strings here.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+}
+
+/// PCM information.
+#[repr(C)]
+#[derive(AsBytes, FromBytes, FromZeroes)]
+pub struct VirtIOSndPcmInfo {
+    hdr: VirtIOSndInfo,
+    features: u32, /* 1 << VIRTIO_SND_PCM_F_XXX */
+    formats: u64,  /* 1 << VIRTIO_SND_PCM_FMT_XXX */
+    rate: u64,     /* 1 << VIRTIO_SND_PCM_RATE_XXX */
```
qwandor, line 1406: Could this be a `PcmRate` rather than a `u64`, to avoid manually converting back and forth? Likewise the other bit flags fields.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+
+impl Into<u64> for PcmFrameRate {
+    fn into(self) -> u64 {
+        1 << self as usize
+    }
+}
+
+/// PCM information.
+#[repr(C)]
+#[derive(AsBytes, FromBytes, FromZeroes)]
```
qwandor, line 1401: Derive `Clone`, `Debug`, `Eq`, `PartialEq`.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    setup: bool,
+    buffer_bytes: u32,
+    period_bytes: u32,
+    features: PcmFeatures,
+    channels: u8,
+    format: PcmFormats,
+    rate: PcmRate,
+}
+
+impl Default for PcmParameters {
```
qwandor, line 1447: Can't this be derived automatically?

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        buf1[..U32_SIZE].copy_from_slice(&stream_id.to_le_bytes());
+        buf1[U32_SIZE..U32_SIZE + buffer_size].copy_from_slice(&frames[..buffer_size]);
+
+        buf2[..U32_SIZE].copy_from_slice(&stream_id.to_le_bytes());
+        buf2[U32_SIZE..U32_SIZE + buffer_size]
+            .copy_from_slice(&frames[buffer_size..buffer_size * 2]);
+
+        let mut outputs = vec![0; 32];
+        let mut token1 = unsafe { self.tx_queue.add(&[&buf1], &mut [&mut outputs]).unwrap() };
+        let mut token2 = unsafe { self.tx_queue.add(&[&buf2], &mut [&mut outputs]).unwrap() };
```
qwandor, line 528: Adding two entries to the queue with the same output buffer is unsound. There must not be any other reference to an output buffer until it is popped.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        assert!(buffer_size * 2 <= frames.len());
+        let mut buf1 = vec![0; U32_SIZE + buffer_size];
+        let mut buf2 = vec![0; U32_SIZE + buffer_size];
+
+        buf1[..U32_SIZE].copy_from_slice(&stream_id.to_le_bytes());
+        buf1[U32_SIZE..U32_SIZE + buffer_size].copy_from_slice(&frames[..buffer_size]);
+
+        buf2[..U32_SIZE].copy_from_slice(&stream_id.to_le_bytes());
+        buf2[U32_SIZE..U32_SIZE + buffer_size]
+            .copy_from_slice(&frames[buffer_size..buffer_size * 2]);
```
qwandor, line 524: You could avoid this extra copy by adding the stream ID and buffer separately to the TX queue. Something like:

```rust
let (first, rest) = frames.split_at(buffer_size);
let token1 = unsafe { self.tx_queue.add(&[&stream_id.to_le_bytes(), first], &mut[&mut output1]) }?;
let token2 = unsafe { self.tx_queue.add(&[&stream_id.to_le_bytes(), &rest[..buffer_size]], &mut[&mut output2]) }?;
```

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+            }
+
+            let start_byte = i * buffer_size;
+            let end_byte = if i != xfer_times - 1 {
+                (i + 1) * buffer_size
+            } else {
+                frames.len()
+            };
+            if turn1 {
+                while let Err(_) = unsafe {
```
qwandor, line 550: Some errors should be passed back to the caller rather than being retried forever. I suggest this loop until `self.tx_queue.can_pop()` returns true instead, and then calls `pop_used` with any error being propagated up.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        } else {
+            frames.len() / buffer_size + 1
+        };
+
+        let mut turn1 = true;
+
+        for i in 2..xfer_times {
+            if self.tx_queue.should_notify() {
+                self.transport.notify(TX_QUEUE_IDX);
+            }
```
qwandor, line 541: This notification needs to happen even if `xfer_times` <= 2. Move it up to just after the `self.tx_queue.add` calls above, then have it again below the add calls below.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        if !self.set_up {
+            self.set_up();
+            self.set_up = true;
+        }
+        if !self.pcm_parameters[stream_id as usize].setup {
+            warn!("Please set parameters for a stream before using it!");
+            return Err(Error::IoError);
+        }
+        let buffer_size = self.pcm_parameters[stream_id as usize].buffer_bytes as usize;
+        assert!(buffer_size * 2 <= frames.len());
```
qwandor, line 515: Why is the caller not allowed to send a smaller amount of data?

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+                spin_loop();
+            }
+        }
+        Ok(())
+    }
+
+    /// Transfer PCM frame to device, based on the stream type(OUTPUT/INPUT).
+    /// 
+    /// This is a non-blocking method that returns a token.
+    pub fn pcm_xfer_nb(&mut self, stream_id: u32, frames: &[u8]) -> Result<u16> {
```
qwandor, line 595: Calling this twice in quick succession will cause undefined behaviour as the same `self.output_rsp` is used for two entries in the virt queue, and the device may be writing to the second while we are reading from the first.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        prepare,
+        Self::Prepare,
+        Self::SetParams | Self::Prepare | Self::Release
+    );
+
+    set_new_state!(start, Self::Start, Self::Prepare | Self::Stop);
+
+    set_new_state!(stop, Self::Stop, Self::Start);
+
+    set_new_state!(release, Self::Release, Self::Prepare | Self::Stop);
```
qwandor, line 791: The methods generated by these macros don't seem to be called anywhere, what are they intended for?

muxinyu1:
Thank you for your advice. I will find time to address these issues!

src/volatile.rs
```diff
@@ -11,12 +11,12 @@ impl<T: Copy> ReadOnly<T> {
 }
 
 /// An MMIO register which can only be written to.
-#[derive(Default)]
+#[derive(Default, Debug, Clone, Copy)]
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/volatile.rs
```diff
@@ -1,5 +1,5 @@
 /// An MMIO register which can only be read from.
-#[derive(Default)]
+#[derive(Default, Debug, Clone, Copy)]
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

Cargo.toml
```diff
@@ -19,6 +19,8 @@ log = "0.4"
 bitflags = "2.3.0"
 zerocopy = { version = "0.7.5", features = ["derive"] }
```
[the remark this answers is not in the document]
muxinyu1: Removed.

src/device/mod.rs
```diff
@@ -12,4 +12,6 @@ pub mod net;
 
 pub mod socket;
 
+pub mod sound;
```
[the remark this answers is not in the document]
muxinyu1: `#[cfg(feature = "alloc")` is added.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    VirtioSndChmapBc,   /* bottom center */
+    VirtioSndChmapBlc,  /* bottom left center */
+    VirtioSndChmapBrc,  /* bottom right center */
+}
+
+/// maximum possible number of channels
+const VIRTIO_SND_CHMAP_MAX_SIZE: usize = 18;
+
+#[repr(C)]
+#[derive(FromBytes, FromZeroes)]
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+            let position = format!("{:?}", ChannelPosition::from(self.positions[i]));
+            positions.push(position);
+        }
+        write!(
+            f,
+            "direction: {}, channels: {}, postions: {:?}",
+            direction, self.channels, positions
+        )
+    }
+}
```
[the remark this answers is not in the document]
muxinyu1: I haven't added unit tests, but I've added demo code for playing audio in `examples/riscv`. Simply execute `make qemu` in `examples/riscv` to play the demo audio.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+
+    event_buf: Box<[u8]>,
+
+    pcm_states: Vec<PCMState>,
+
+    token_buf: BTreeMap<u16, Box<Vec<u8>>>, // store token and its input buf
+}
+
+impl<H: Hal, T: Transport> VirtIOSound<H, T> {
+    /// Craete a new VirtIO-Sound driver.
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    pub fn new(mut transport: T) -> Result<Self> {
+        let negotiated_features = transport.begin_init(SUPPORTED_FEATURES);
+        info!(
+            "[sound device] negotiated_features: {}",
+            String::from(negotiated_features)
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    set_new_state!(release, Self::Release, Self::Prepare | Self::Stop);
+}
+
+const QUEUE_SIZE: u16 = 32;
+const CONTROL_QUEUE_IDX: u16 = 0;
+const EVENT_QUEUE_IDX: u16 = 1;
+const TX_QUEUE_IDX: u16 = 2;
+const RX_QUEUE_IDX: u16 = 3;
+
+const SUPPORTED_FEATURES: SoundFeatures = SoundFeatures::all();
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+
+bitflags! {
+    /// In virtIO v1.2, there are no specific features defined for virtio-sound, so now it's common
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        );
+
+        let control_queue = VirtQueue::new(
+            &mut transport,
+            CONTROL_QUEUE_IDX,
+            negotiated_features.contains(SoundFeatures::VIRTIO_F_INDIRECT_DESC),
+            negotiated_features.contains(SoundFeatures::VIRTIO_F_EVENT_IDX),
+        )?;
+        // The driver MUST populate the event queue
+        // with empty buffers of at least the struct virtio_snd_event size(struct VirtIOSndEvent Size in config.rs)
```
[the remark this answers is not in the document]
muxinyu1: Sorry, that comment was from before and I forgot to delete it. It's been removed now.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        let rx_queue = VirtQueue::new(
+            &mut transport,
+            RX_QUEUE_IDX,
+            negotiated_features.contains(SoundFeatures::VIRTIO_F_INDIRECT_DESC),
+            negotiated_features.contains(SoundFeatures::VIRTIO_F_EVENT_IDX),
+        )?;
+
+        // read configuration space
+        let config_ptr = transport.config_space::<VirtIOSoundConfig>()?;
+        let (jacks, streams, chmaps) = unsafe {
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        self.chmaps
+    }
+
+    /// Acknowledge interrupt.
+    pub fn ack_interrupt(&mut self) -> bool {
+        self.transport.ack_interrupt()
+    }
+
+    fn request<Req: AsBytes, Rsp: FromBytes>(&mut self, req: Req) -> Result<Rsp> {
+        req.write_to_prefix(&mut self.queue_buf_send).unwrap();
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+            error!("jack_start_id + jack_count > jacks! There are not enough jacks to be queried!");
+            return Err(Error::IoError);
+        }
+        let hdr: VirtIOSndHdr = self.request(VirtIOSndQueryInfo {
+            hdr: ItemInfomationRequestType::VirtioSndRJackInfo.into(),
+            start_id: jack_start_id,
+            count: jack_count,
+            size: mem::size_of::<VirtIOSndJackInfo>() as u32,
+        })?;
+        // mem::size_of::<VirtIOSndHdr>() bytes are header
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        if let Ok(chmap_infos) = self.chmap_info(0, self.chmaps) {
+            for chmap_info in &chmap_infos {
+                info!("[sound device] chmap_info: {}", chmap_info);
+            }
+            self.chmap_infos = Some(chmap_infos);
+        } else {
+            self.chmap_infos = Some(vec![]);
+            warn!("[sound device] There are no available chmaps!");
+        }
+        // set pcm state to defalut
```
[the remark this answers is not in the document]
muxinyu1: Resolved.

src/device/sound.rs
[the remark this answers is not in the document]
muxinyu1: Now the driver supports only `Feature::RING_EVENT_IDX`.

src/device/sound.rs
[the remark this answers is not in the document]
muxinyu1: I replaced `SoundFeature` with `device::common::Feature`.

src/device/sound.rs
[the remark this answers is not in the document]
muxinyu1: Safety comment is added.

src/device/sound.rs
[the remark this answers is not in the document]
muxinyu1: Removed the extra copy.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+            pcm_infos.push(pcm_info);
+        }
+        Ok(pcm_infos)
+    }
+
+    /// Query information about the available chmaps.
+    fn chmap_info(
+        &mut self,
+        chmaps_start_id: u32,
+        chamaps_count: u32,
```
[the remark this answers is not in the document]
muxinyu1: Fixed.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+            self.set_up();
+            self.set_up = true;
+        }
+        if self.jacks == 0 {
+            error!("[sound device] There is no available jacks!");
+            return Err(Error::InvalidParam);
+        }
+        if self.jack_infos == None {
+            error!("Could not read jack_infos, you need to call set_up() to initate it.");
+            return Err(Error::InvalidParam);
```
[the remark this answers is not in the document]
muxinyu1: Fixed.

src/device/sound.rs
muxinyu1: In the `VirtIOSndJackInfo` struct, there are several bitflags represented as u32 (or u64) according to the virtio v1.2 specification. Directly implementing `Debug` for `VirtIOSndJackInfo` cannot print detailed information about these bitflags; it only prints their u32 value. Therefore, it is necessary to use `bitflags::parser::to_writer` to format the u32 bitflags for clearer output.

src/device/sound.rs
muxinyu1: Added.

src/device/sound.rs
muxinyu1: Added.

src/device/sound.rs
muxinyu1: Removed.

src/device/sound.rs
muxinyu1: Yes, `ItemInformationRequestType` is indeed a subset of `CommandCode`. Unlike `CommandCode`, `ItemInformationRequestType` is public. There are some request codes in `CommandCode` that users do not need. Therefore, for user convenience, I selected its subset and defined it as a public enumeration.

src/device/sound.rs
muxinyu1: I believe we should not convert u32, u64, and so on into bitflags for these structure fields. The naming and types of these fields are copied directly from the C implementation in Virtio v1.2. Therefore, I think we should not modify these standard-defined structures.

src/device/sound.rs
muxinyu1: Added.

src/device/sound.rs
muxinyu1: Fixed, now this is derived automatically.

src/device/sound.rs
muxinyu1: I'm not going to do anything with this `outputs` here; it merely serves the purpose of receiving return values sent by the device.

src/device/sound.rs
muxinyu1: Now I've removed `buf1` and `buf2` to avoid the extra copy.

muxinyu1: Now I've removed `buf1` and `buf2` to avoid the extra copy.

src/device/sound.rs
muxinyu1: Fixed.

src/device/sound.rs
muxinyu1: Now I've removed the unused macros.

src/device/sound.rs
muxinyu1: Now I've fixed this bug, and now caller is allowed to send a smaller data.

src/device/sound.rs
muxinyu1: Fixed.

muxinyu1:
I have made some changes to the driver code based on your suggestions. If you have time, please review them.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+    /// indicates the current jack connection status (1 - connected, 0 - disconnected)
+    connected: u8,
+
+    _padding: [u8; 7],
+}
+
+impl Display for VirtIOSndJackInfo {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        let jack_features = JackFeatures::from(self.features);
+        let mut jack_features_str = String::new();
```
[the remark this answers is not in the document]
qwandor: I've done this myself now and pushed to your branch.

src/device/sound.rs
qwandor: Done this myself.

src/device/sound.rs
qwandor: Do you actually use `pcm_states` anywhere?

src/device/sound.rs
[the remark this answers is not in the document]
qwandor: I've added a test for reading the config space. Some more unit tests of functionality would be good, though not required for merging.

examples/riscv/music_44100Hz_u8_stereo.raw
qwandor, line 1: Please use a shorter example sound file. This makes CI take ages to run the example, and also makes the git repository much bigger than it needs to be.

src/device/sound.rs
muxinyu1: **Other places indeed do not use `pcm_states`**. These macros were originally intended to determine if the caller switched the PCM states in the correct order. However, due to time constraints, this state checking, which could have enhanced robustness, was not implemented.

src/device/sound.rs
```diff
@@ -0,0 +1,1563 @@
...
+        if stream_id >= self.pcm_infos.as_ref().unwrap().len() as u32 {
+            return Err(Error::InvalidParam);
+        }
+        let pcm_info = &self.pcm_infos.as_ref().unwrap()[stream_id as usize];
+        Ok(pcm_info.features.into())
+    }
+
+    /// Get the latest notification.
+    pub fn latest_notification(&mut self) -> Result<Notification> {
+        if let Some(token) = self.event_queue.peek_used() {
```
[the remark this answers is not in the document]
muxinyu1: I removed this function because I couldn't understand the corresponding section in the virtio v1.2 manual about 'the device notifying the driver.' According to the manual, the device should actively send notifications to the event_queue, but my tests showed that it wasn't sending any notifications. Perhaps my implementation of this notification mechanism has some issues.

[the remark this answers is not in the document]
qwandor: Did you try adding buffers to the event queue for it to write to? It's always up to the driver to allocate buffers, devices can't just add them to the queue out of nowhere.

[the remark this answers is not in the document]
muxinyu1: When should I add buffers to the event queue? Should I do it when calling the `latest_notification` function, or during driver initialization? If I add buffers during initialization, how many should I add? I'm not quite clear on this.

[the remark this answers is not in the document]
qwandor: I'd suggest just adding a single buffer to the event queue during initialisation. `latest_notification` can then add the buffer back to the queue after reading from it.

examples/riscv/music_44100Hz_u8_stereo.raw
qwandor: You don't seem to have pushed this change.

src/device/sound.rs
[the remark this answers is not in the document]
qwandor: Filling the queue with buffers is also a reasonable option. I've sent #149 with a safe abstraction which may be useful for this.

examples/riscv/music_44100Hz_u8_stereo.raw
muxinyu1: Now pushed.

src/device/sound.rs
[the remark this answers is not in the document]
muxinyu1: I added some buffers to the `event_queue` for the device to write into and notified the device in `set_up()`, but when calling `latest_notification()`, it seems the device still hasn't written notifications to the previously added buffers. Is there an issue with my implementation?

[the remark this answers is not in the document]
muxinyu1: If the implementation to get device notifications is too difficult to grasp, I think it's okay not to implement it. As shown in this [PPT](https://fosdem.org/2024/events/attachments/fosdem-2024-1910-making-virtio-sing-implementing-virtio-sound-in-rust-vmm-project/slides/22175/implementing_virtio-sound_in_rust-vmm_project_HhqbjPY.pdf), they also did not implement an event_queue.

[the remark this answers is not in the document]
qwandor: I've changed this to use `OwningQueue` to manage the buffers.

muxinyu1:
Thank you very much for your help on my PR! If there is anything where I can help, I will do my best to provide assistance!

src/device/sound.rs
qwandor: I've rewritten this to fix the issue.

qwandor (approved):

PR #139, opened 2024-06-05
Make vsock buffer sizes configurable

qwandor:


src/device/socket/connectionmanager.rs
```diff
@@ -56,24 +64,36 @@ struct Connection {
...
+    /// the given per-connection buffer capacity.
+    pub fn new_with_capacity(
+        driver: VirtIOSocket<H, T, RX_BUFFER_SIZE>,
+        per_connection_buffer_capacity: usize,
+    ) -> Self {
         Self {
             driver,
             connections: Vec::new(),
             listening_ports: Vec::new(),
+            per_connection_buffer_capacity,
```
thedataking, line 96: Are there any requirements on `per_connection_buffer_capacity` similar to `RX_BUFFER_SIZE`? 

If so, would it make sense to validate the buffer size here or would an invalid buffer fail a check elsewhere?

thedataking (approved):
Tested with a buffer size of 4096 for both `PER_CONNETION_BUFFER_CAPACITY` and `RX_BUFFER_SIZE`. Not sure if per connection buffer sizes need validation, otherwise this looks great to me!

src/device/socket/connectionmanager.rs
qwandor: As far as I can tell the only requirement is that it fit in a u32, so I've changed the type to that.

PR #144, opened 2024-06-12
Add typed getters for input device config space

qwandor:
Rather than just adding bytes to a buffer, these return the appropriate type for the selector.

The size of the data available for a particular selector is given by the device, so only read that much data rather than the full 128 bytes every time.

### DIFF
--- a/src/device/input.rs
+++ b/src/device/input.rs
@@ -5,9 +5,13 @@ use crate::hal::Hal;
 use crate::queue::VirtQueue;
 use crate::transport::Transport;
 use crate::volatile::{volread, volwrite, ReadOnly, VolatileReadable, WriteOnly};
-use crate::Result;
-use alloc::boxed::Box;
+use crate::Error;
+use alloc::{
+    boxed::Box,
+    string::{FromUtf8Error, String},
+};
 use core::cmp::min;
+use core::mem::size_of;
 use core::ptr::{addr_of, NonNull};
 use zerocopy::{AsBytes, FromBytes, FromZeroes};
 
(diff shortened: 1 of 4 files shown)


src/device/input.rs
```diff
@@ -107,17 +112,90 @@ impl<H: Hal, T: Transport> VirtIOInput<H, T> {
...
         let size;
-        let data;
         // Safe because config points to a valid MMIO region for the config space.
         unsafe {
             volwrite!(self.config, select, select as u8);
             volwrite!(self.config, subsel, subsel);
             size = volread!(self.config, size);
-            data = volread!(self.config, data);
+            let size_to_copy = min(usize::from(size), out.len());
+            for i in 0..size_to_copy {
```
equation314, line 121: Fix clippy warning

src/device/input.rs
```diff
@@ -107,17 +112,90 @@ impl<H: Hal, T: Transport> VirtIOInput<H, T> {
...
     }
+
+    /// Queries a specific piece of information by `select` and `subsel`, allocates a sufficiently
+    /// large byte buffer for it, and returns it.
+    fn query_config_select_alloc(&mut self, select: InputConfigSelect, subsel: u8) -> Box<[u8]> {
+        // Safe because config points to a valid MMIO region for the config space.
+        unsafe {
+            volwrite!(self.config, select, select as u8);
+            volwrite!(self.config, subsel, subsel);
+            let size = usize::from(volread!(self.config, size));
```
equation314, line 135: We need to check if `size` is valid here.

src/device/input.rs
qwandor: Done.

src/device/input.rs
qwandor: Done.

PR #153, opened 2024-07-30
Use `flat_device_tree` crate for examples

qwandor:
The `fdt` crate hasn't been updated in a while, `flat_device_tree` is a fork with an improved API. I also fixed a bug which got the prefetchable attribute for PCI ranges from the wrong bit.


equation314 (approved):
LGTM

PR #164, opened 2024-10-29
Add hypercall-based PCI CAM and transport for x86-64 pKVM.

qwandor:
VMs running under pKVM on x86-64 cannot do MMIO, so a hypercall based transport is provided instead.

### DIFF
--- a/src/transport/x86_64.rs
+++ b/src/transport/x86_64.rs
@@ -1,5 +1,7 @@
 //! x86-64 specific transports.
 
+mod hypercalls;
+
 use super::{
     pci::{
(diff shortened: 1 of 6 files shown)


jaszczyk-grzegorz (approved):

src/transport/x86_64.rs
```diff
@@ -0,0 +1,124 @@
+//! x86-64 specific transports.
```
jaszczyk-grzegorz, line 1: nit: maybe "hypercall based, x86-64 specific transport"?

jaszczyk-grzegorz (approved):
LGTM and thanks for preparing this PR!

PR #171, opened 2024-12-03
add macro to read config fields

fkm3:
Ensures that the offset and type match.

Proof of concept of https://github.com/rcore-os/virtio-drivers/pull/163#discussion_r1859185821

This doesn't mesh well with the existing config struct definitions, which use
types to specify qualities like "read only" and "volatile". Since
https://github.com/rcore-os/virtio-drivers/pull/163, it seems like those are
essentially ignored, so, if this macro seems like a good idea, it would
probably make sense to change the config types to be plain-old-data struct that
implement zerocopy traits.

### DIFF
--- a/src/device/gpu.rs
+++ b/src/device/gpu.rs
@@ -2,6 +2,7 @@
 
 use crate::hal::{BufferDirection, Dma, Hal};
 use crate::queue::VirtQueue;
+use crate::transport::read_config_field;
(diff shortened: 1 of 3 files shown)


qwandor:
What do you think of the approach in #170?

fkm3:
looks good, closing this one

Issue #177, opened 2025-02-27
PCI 64-bit memory BAR size should not be limited to 32 bits


danielverkamp:
64-bit-wide PCI base address registers have a 64-bit size as defined by the PCI spec, but the current enumeration code assumes that the size fits in 32 bits (`BarInfo::Memory` has a `size` field that is only `u32`, not `u64`). This means that the size of BARs larger than 2 GB will be incorrectly truncated (treated as 0, which is interpreted as not present elsewhere in the code). One practical case where this may happen is a virtio-gpu device with a shared memory region, which can easily be 4 GB or larger (in qemu, this is controlled by the "hostmem" property).

A related issue is `struct VirtioCapabilityInfo`, which represents the offset and length of a region within a BAR; this structure also has only 32-bit-wide `offset` and `length` fields, but these may need to represent 64-bit quantities when `virtio_pci_cap64` is used, in particular as required for shared memory regions in the virtio PCI transport.

qwandor:
Good point. Could you send a PR to fix this too?

danielverkamp:
Yes, I have a patch locally and am working on exercising it on different VMMs.

I also don't know what kind of API stability this crate wants to provide; this change will modify the `pub size` field's type, potentially requiring changes in code that depends on this crate, but since it is still version 0.x, that's probably fine.

PR #178, opened 2025-02-28
PCI: read full 64-bit size of 64-bit memory BARs

danielverkamp:
64-bit memory BARs not only have a 64-bit-wide address, but also a 64-bit-wide size. Change the bar_info() function to retrieve the upper 32 bits of the size and include it in the size calculation.

This modifies the return type of the BarInfo::memory_address_size() function, so callers may need to modify their code to adapt to the new 64-bit size.

Fixes GitHub issue #177.

### DIFF
--- a/examples/aarch64/src/main.rs
+++ b/examples/aarch64/src/main.rs
@@ -418,6 +418,13 @@ fn allocate_bars(
             address_type, size, ..
         } = info
         {
+            // For now, only attempt to allocate 32-bit memory regions.
+            if size > u32::MAX.into() {
+                warn!("Skipping BAR {} with size {:#x}", bar_index, size);
+                continue;
+            }
+            let size = size as u32;
+
             match address_type {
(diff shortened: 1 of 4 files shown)


qwandor (approved):
Looks good, thanks!

danielverkamp:
Sorry, the build failure should be fixed now (I thought I tested all the architectures, but I must have skipped aarch64 on the last pass).

danielverkamp:
Is that for 32-bit or 64-bit BARs? With crosvm, qemu, something else?

I will take a look and see if I can reproduce locally - the math is a bit tricky and could probably stand to be simplified anyway.

qwandor:
The bug shows up with both crosvm and QEMU. I've added a unit test in #191 which catches the issue. I also noticed a bug where IO BAR sizes less than 16 bytes were read incorrectly, which I fixed.

PR #179, opened 2025-03-05
Change `// Safe because` comments to proper `// SAFETY:` comments `clippy` recognizes

kkysen:
Most of the `unsafe` blocks already have safety comments, but they were in a non-standard `// Safe because` format that `clippy::undocumented_unsafe_blocks` doesn't understand, rather than `// SAFETY:` comments.  This just switches them so that `clippy` understands them and can check them with `#![deny(clippy::undocumented_unsafe_blocks)]`.  This brings down the number of undocumented `unsafe`s, according to `clippy` and excluding `#[cfg(test)]` code, from 69 to 15.


qwandor (approved):
Thanks!

PR #187, opened 2025-03-12
Add remaining safety comments

randomPoison:
Add `SAFETY` comments to the remaining undocumented `unsafe` blocks. I did my best to accurately cover why each block is unsafe, but I'm not familiar with virtio and may have missed some subtle details. Let me know if anything needs adjustment.

### DIFF
--- a/src/device/gpu.rs
+++ b/src/device/gpu.rs
@@ -115,12 +115,13 @@ impl<H: Hal, T: Transport> VirtIOGpu<H, T> {
         // map frame buffer to screen
         self.set_scanout(display_info.rect, SCANOUT_ID, RESOURCE_ID_FB)?;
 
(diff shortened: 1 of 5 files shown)


randomPoison:
Oh, also when this lands it might be a good idea to add `#![warn(clippy::undocumented_unsafe_blocks)]` (or deny them outright). I can add that to this PR if desired.

qwandor:
Thanks for doing this!

> Oh, also when this lands it might be a good idea to add `#![warn(clippy::undocumented_unsafe_blocks)]` (or deny them outright). I can add that to this PR if desired.

Good idea, please do. Let's make it `deny` rather than `warn` if possible.

randomPoison:
Okay, added `clippy::undocumented_unsafe_blocks` to the existing `#![deny()]` attribute.

qwandor (approved):

PR #190, opened 2025-04-17
Add VirtIORng driver

00xc:
Add a driver for the entropy source device, also known as virtio-rng in qemu. Add basic usage as well in the repo examples for all architectures.

### DIFF
--- a/src/device/rng.rs
+++ b/src/device/rng.rs
@@ -1,0 +1,58 @@
+//! Driver for VirtIO random number generator devices.
+use super::common::Feature;
+use crate::{queue::VirtQueue, transport::Transport, Hal, Result};
+
+// VirtioRNG only uses one queue
+const QUEUE_IDX: u16 = 0;
+const QUEUE_SIZE: usize = 8;
+const SUPPORTED_FEATURES: Feature = Feature::RING_INDIRECT_DESC.union(Feature::RING_EVENT_IDX);
+
+/// Driver for a VirtIO random number generator device.
+pub struct VirtIORng<H: Hal, T: Transport> {
+    transport: T,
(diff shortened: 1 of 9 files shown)


src/device/rng.rs
```diff
@@ -0,0 +1,56 @@
...
+use crate::{queue::VirtQueue, transport::Transport, Hal, Result};
+
+// VirtioRNG only uses one queue
+const QUEUE_IDX: u16 = 0;
+const QUEUE_SIZE: usize = 8;
+
+bitflags::bitflags! {
+    #[derive(Clone, Copy, Debug, Default)]
+    struct RngFeature: u64 {}
+}
```
qwandor, line 11: If there are no device-specific features then use `device::common::Feature`.

src/device/rng.rs
```diff
@@ -0,0 +1,56 @@
...
+/// Driver for a VirtIO random number generator device.
+pub struct VirtIORng<H: Hal, T: Transport> {
+    transport: T,
+    queue: VirtQueue<H, QUEUE_SIZE>,
+}
+
+impl<H: Hal, T: Transport> VirtIORng<H, T> {
+    /// Create a new driver with the given transport.
+    pub fn new(mut transport: T) -> Result<Self> {
+        transport.begin_init(RngFeature::empty());
```
qwandor, line 22: It should be simple to add support for the common `RING_EVENT_IDX` and `RING_INDIRECT_DESC` features. See `VirtIOInput::new` for an example of how to negotiate them and configure the queue appropriately.

qwandor (commented):
Thanks! Looks good, just some minor comments about feature negotiation.

A test would also be nice, but optional.

qwandor (approved):
Thanks!

PR #191, opened 2025-05-06
Fix sizing of IO BARs and unused BARs

qwandor:


src/transport/pci/bus.rs
```diff
@@ -695,3 +699,216 @@ impl From<u8> for HeaderType {
...
+                0xffffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xffffffff,
+            ],
+            status_command: 0,
+        };
+        let mut root = PciRoot::new(fake_cam);
+
+        assert_eq!(
+            root.bars(device_function).unwrap(),
+            [
+                Some(BarInfo::Memory {
```
danielverkamp, line 727: Maybe something for a separate API change, but it seems like it would be clearer to represent unimplemented BARs as `None` instead of `Some(..., size: 0, ...`.

danielverkamp (approved):
Thanks, the fix for the unimplemented register case (`size_mask == 0`) looks correct to me.

src/transport/pci/bus.rs
```diff
@@ -695,3 +699,216 @@ impl From<u8> for HeaderType {
...
+                self.status_command
+            } else if register_offset >= BAR0_OFFSET && register_offset < 0x28 {
+                let bar_index = usize::from((register_offset - BAR0_OFFSET) / 4);
+                println!(
+                    "Reading {} from bar {}",
+                    self.bar_values[bar_index], bar_index
+                );
+                self.bar_values[bar_index]
+            } else {
+                return 0xffffffff;
```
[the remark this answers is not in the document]
qwandor: Quite right, this was left over after some refactoring. Fixed now.

src/transport/pci/bus.rs
qwandor: Is it possible to have unimplemented IO BARs or 64-bit memory BARs? The current API allows them to be distinguished, but I'm not sure whether that is actually relevant.

danielverkamp: An unimplemented base address register returns all 0 bits regardless of what is written to it, which just happens to correspond to the bit pattern for a memory-type BAR. So no, there is no "unimplemented IO BAR" or really "unimplemented <type> BAR" of any kind, and returning `BarInfo::Memory` for such a BAR is not really accurate. (Hopefully nobody is relying on the exact API for unused BARs anyway, but it would technically be a breaking API change.)

qwandor: Done in #192, PTAL.

PR #192, opened 2025-05-07
Return None rather than size 0 for unused BARs

qwandor:
This changes the return type of `PciRoot::bar_info`, so is a breaking change.

### DIFF
--- a/src/transport/pci/bus.rs
+++ b/src/transport/pci/bus.rs
@@ -202,8 +202,12 @@ impl<C: ConfigurationAccess> PciRoot<C> {
         let mut bar_index = 0;
         while bar_index < 6 {
             let info = self.bar_info(device_function, bar_index)?;
-            let takes_two_entries = info.takes_two_entries();
-            bars[usize::from(bar_index)] = Some(info);
+            let takes_two_entries = if let Some(info) = &info {
+                info.takes_two_entries()
+            } else {
+                false
+            };
+            bars[usize::from(bar_index)] = info;
             bar_index += if takes_two_entries { 2 } else { 1 };
         }
         Ok(bars)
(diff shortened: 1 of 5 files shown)


src/transport/pci/bus.rs
```diff
@@ -238,6 +242,10 @@ impl<C: ConfigurationAccess> PciRoot<C> {
                 .read_word(device_function, BAR0_OFFSET + 4 * bar_index),
         );
 
+        if size_mask == 0 {
+            return Ok(None);
```
danielverkamp, line 246: We still need to restore the original BAR value and command register configuration, which returning early skips. Maybe the "Restore the original value" step could be moved up before this early return? But the command register IO/mem decode disabling + re-enabling still needs to be wrapped around all BAR modifications, including the writing of the high half of 64-bit BARs below, so it's a bit trickier to rearrange (unless you just want to duplicate the `self.set_command()` call in the early return branch).

It would probably be good to verify that everything is put back the way it was originally in the unit tests as well, so this kind of thing would be caught.

src/transport/pci/bus.rs
```diff
@@ -202,8 +202,12 @@ impl<C: ConfigurationAccess> PciRoot<C> {
         let mut bar_index = 0;
         while bar_index < 6 {
             let info = self.bar_info(device_function, bar_index)?;
-            let takes_two_entries = info.takes_two_entries();
-            bars[usize::from(bar_index)] = Some(info);
+            let takes_two_entries = if let Some(info) = &info {
```
danielverkamp, line 205: nit: consider something like:

```
let bar_entries = if info.is_some_and(BarInfo::info.takes_two_entries) { 2 } else { 1 };
...
bar_index += bar_entries;
```

danielverkamp (commented):
The API change looks good to me; a few additional notes below.

src/transport/pci/bus.rs
danielverkamp: (I guess restoring the original BAR value doesn't really matter, since it should be all 0 bits in this case, but the point about the command register still stands.)

src/transport/pci/bus.rs
qwandor: Done.

danielverkamp (approved):

Issue #201, opened 2025-07-02
Would it be feasible to introduce a generic virtio device type?


ken4647:
In some cases, virtio is simply used as one possible kind of communication channels, such as with virtio-9p. For virtio-9p, it isn't really appropriate to place the core logic inside this crate, because 9PFS can also use other underlying transports like TCP. In terms of transport functionality, the code is too simple and not inherently tied to 9P itself. 

A more suitable design would be to treat virtio and TCP as transport layers, while 9P serves as a separate logical abstraction. This approach would help avoid redundant implementations.

```
|----------------------------|
|         9P (Client)        | ←→ 9P Protocol Layer
|----------------------------|
            ↑
|---------------------------|
|   Transport Abstraction   | ←→ Generic Transport Interface
|---------------------------|
    ↑                 ↑
|--------|    |-------------|
| virtio |    |     TCP     |  ←→ Transport Implementations
|--------|    |-------------|
```
In addition, some customized virtio devices may also represent potential demand.

PR #202, opened 2025-07-02
feature: add generic device named as utility

ken4647:
Here is a possible demonstration of the effect. The purpose of implementing this interface is to expose sufficient low-level space to developers, in order to avoid code duplication and waste. The main examples are 9p and customized virtio devices.

For simplicity, it assume here that there is only one virtio queue. Of course, I think the naming of "utility" still has room for discussion.

Related to #201 .


qwandor (commented):
Rather than adding a generic driver like this, which has assumptions such as only having one virtqueue, how would you feel about making `VirtQueue` and any other types which you need for your driver public, so that you can implement the driver outside of this crate?

PR #203, opened 2025-07-11
Make `ack_interrupt` return interrupt status

thedataking:
This appears to be a known limitation in the current implementation. Without this change, interrupt service routines are unable to tell why they are running (is there a new virtqueue packet to process or did the device confirmation change).

This PR also makes a tiny cleanup change to remove a safety comment that is no longer needed.

### DIFF
--- a/src/transport/mod.rs
+++ b/src/transport/mod.rs
@@ -69,7 +69,7 @@ pub trait Transport {
     /// Acknowledges an interrupt.
     ///
     /// Returns true on success.
-    fn ack_interrupt(&mut self) -> bool;
+    fn ack_interrupt(&mut self) -> InterruptStatus;
 
     /// Begins initializing the device.
     ///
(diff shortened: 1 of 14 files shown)


src/transport/mod.rs
```diff
@@ -175,6 +175,23 @@ bitflags! {
     }
 }
 
+/// ISR status field. Reading this register resets it to 0 and causes the device
+/// to de-assert the interrupt.
```
qwandor, line 179: I think this comment belongs on the field rather than the type. This type is part of the public API of all the drivers, so it makes more sense to describe it from that point of view. E.g. something like "The set of interrupts which were pending".

src/transport/mod.rs
```diff
@@ -175,6 +175,23 @@ bitflags! {
...
+/// Ref: 4.1.4.5 ISR status capability
+#[derive(Copy, Clone, Default, Eq, FromBytes, PartialEq)]
+pub struct InterruptStatus(u32);
+
+bitflags! {
+    impl InterruptStatus: u32 {
+        /// Indicates that a virtqueue buffer was used
+        const QUEUE_INTERRUPT = 1;
+
+        /// Indicates that a virtio PCI device changed its configuration state
```
qwandor, line 190: Is this actually specific to PCI devices, or does it also apply to VirtIO devices with other transports?

qwandor: Section 4.2.2 MMIO Device Register Layout also mentions a "configuration change notification" bit in the InterruptStatus register.

src/transport/mod.rs
```diff
@@ -175,6 +175,22 @@ bitflags! {
...
+/// The set of interrupts which were pending
+///
+/// Ref: 4.1.4.5 ISR status capability
+#[derive(Copy, Clone, Default, Eq, FromBytes, PartialEq)]
+pub struct InterruptStatus(u32);
+
+bitflags! {
+    impl InterruptStatus: u32 {
+        /// Indicates that a virtqueue buffer was used
+        const QUEUE_INTERRUPT = 1;
```
qwandor, line 187: Let's make this `1 << 0` to make it clear that it's a bitfield.

src/transport/mod.rs
```diff
@@ -175,6 +175,22 @@ bitflags! {
...
+
+        /// Indicates that a virtio PCI device changed its configuration state
+        const DEVICE_CONFIGURATION_INTERRUPT = 2;
```
qwandor, line 190: Likewise `1 << 1` here.

PR #205, opened 2025-09-16
Use u64 rather than usize for PhysAddr

qwandor:
On 32-bit platforms VirtIO still uses 64-bit physical addresses.


qwandor:
This should help with #204.

PR #244, opened 2026-05-17
Add rtc driver

epilys:
Can be used with rust-vmm backend:

  https://github.com/rust-vmm/vhost-device/tree/main/vhost-device-rtc

### DIFF
--- a/src/device/rtc.rs
+++ b/src/device/rtc.rs
@@ -1,0 +1,412 @@
+//! Driver for VirtIO RTC devices.
+
+use core::num::NonZeroU64;
+
+use crate::hal::Hal;
+use crate::queue::VirtQueue;
+use crate::transport::Transport;
+use crate::{Error, Result};
+use bitflags::Flags;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+const QUEUE_REQUEST: u16 = 0;
+const QUEUE_ALARM: u16 = 1;
+const SUPPORTED_FEATURES: Feature = Feature::RING_EVENT_IDX
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::VERSION_1)
+    .union(Feature::ACCESS_PLATFORM);
+
+/// Driver for RTC devices based on VIRTIO 1.4.
+pub struct VirtIORtc<H: Hal, T: Transport> {
+    transport: T,
+    negotiated_features: Feature,
+    request_queue: VirtQueue<H, 8>,
+}
+
+impl<H: Hal, T: Transport> VirtIORtc<H, T> {
+    /// Create a new VirtIO-RTC driver.
+    pub fn new(mut transport: T) -> Result<Self> {
+        let negotiated_features = transport.begin_init(SUPPORTED_FEATURES);
+
+        let request_queue = VirtQueue::new(
+            &mut transport,
+            QUEUE_REQUEST,
+            negotiated_features.contains(Feature::RING_INDIRECT_DESC),
+            negotiated_features.contains(Feature::RING_EVENT_IDX),
+            negotiated_features.contains(Feature::ACCESS_PLATFORM),
+        )?;
+        transport.finish_init();
+
+        Ok(Self {
+            transport,
+            negotiated_features,
+            request_queue,
+        })
+    }
+
+    /// Return any unrecognized feature flag bits.
+    ///
+    /// Any bits that cannot be converted to a known device feature will be returned.
+    pub fn invalid_feature_bits(&mut self) -> Option<NonZeroU64> {
+        let device_feature_bits = self.transport.read_device_features();
+        let device_features: Feature = Feature::from_bits_retain(device_feature_bits);
+        NonZeroU64::new(device_features.unknown_bits())
+    }
+
+    /// Discovers the number of clocks.
+    pub fn num_clocks(&mut self) -> Result<u16> {
+        let req = VirtioRtcReqHead {
+            msg_type: VIRTIO_RTC_REQ_CFG,
+            ..VirtioRtcReqHead::default()
+        };
+        let resp: VirtioRtcRespCfg = self.request(req)?;
+        Ok(resp.num_clocks)
+    }
+
+    /// Returns clock capabilities for a specific `clock_id`.
+    pub fn clock_cap(&mut self, clock_id: u16) -> Result<ClockCapabilities> {
+        let req = VirtioRtcReqClockCap {
+            head: VirtioRtcReqHead {
+                msg_type: VIRTIO_RTC_REQ_CLOCK_CAP,
+                ..VirtioRtcReqHead::default()
+            },
+            clock_id,
+            ..VirtioRtcReqClockCap::default()
+        };
+        let resp: VirtioRtcRespClockCap = self.request(req)?;
+        let kind = match resp.type_ {
+            VIRTIO_RTC_CLOCK_UTC => ClockType::Utc,
+            VIRTIO_RTC_CLOCK_TAI => ClockType::Tai,
+            VIRTIO_RTC_CLOCK_MONOTONIC => ClockType::Monotonic,
+            VIRTIO_RTC_CLOCK_UTC_SMEARED => ClockType::UtcSmeared,
+            VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED => ClockType::UtcMaybeSmeared,
+            _ => return Err(Error::Unsupported),
+        };
+        let leap_second_smearing = if kind == ClockType::UtcSmeared {
+            match resp.leap_second_smearing {
+                VIRTIO_RTC_SMEAR_UNSPECIFIED => None,
+                VIRTIO_RTC_SMEAR_UTC_SLS => Some(SmearingVariant::UtcSls),
+                VIRTIO_RTC_SMEAR_NOON_LINEAR => Some(SmearingVariant::NoonLinear),
+                _ => return Err(Error::Unsupported),
+            }
+        } else {
+            None
+        };
+        Ok(ClockCapabilities {
+            kind,
+            leap_second_smearing,
+            alarm_capability: resp.flags & VIRTIO_RTC_FLAG_ALARM_CAP != 0,
+        })
+    }
+
+    /// Read a clock's timestamp.
+    pub fn read(&mut self, clock_id: u16) -> Result<u64> {
+        let req = VirtioRtcReqRead {
+            head: VirtioRtcReqHead {
+                msg_type: VIRTIO_RTC_REQ_READ,
+                ..VirtioRtcReqHead::default()
+            },
+            clock_id,
+            ..VirtioRtcReqRead::default()
+        };
+        let resp: VirtioRtcRespRead = self.request(req)?;
+        Ok(resp.clock_reading)
+    }
+
+    /// Helper method to send a request to the device and block for a response.
+    fn request<Req: IntoBytes + Immutable, Rsp: FromBytes + Default + Immutable + IntoBytes>(
+        &mut self,
+        req: Req,
+    ) -> Result<Rsp> {
+        let mut resp = Rsp::new_zeroed();
+        self.request_queue.add_notify_wait_pop(
+            &[req.as_bytes()],
+            &mut [resp.as_mut_bytes()],
+            &mut self.transport,
+        )?;
+        let head = VirtioRtcRespHead::read_from_prefix(resp.as_bytes())
+            .unwrap()
+            .0;
+        match head.status {
+            VIRTIO_RTC_S_OK => Ok(resp),
+            VIRTIO_RTC_S_EOPNOTSUPP => Err(Error::Unsupported),
+            VIRTIO_RTC_S_ENODEV | VIRTIO_RTC_S_EINVAL => Err(Error::InvalidParam),
+            VIRTIO_RTC_S_EIO => Err(Error::IoError),
+            other => {
+                log::error!("Invalid status response {other:#x}");
+                Err(Error::IoError)
+            }
+        }
+    }
+}
+
+impl<H: Hal, T: Transport> Drop for VirtIORtc<H, T> {
+    fn drop(&mut self) {
+        // Clear any pointers pointing to DMA regions, so the device doesn't try to access them
+        // after they have been freed.
+        self.transport.queue_unset(QUEUE_REQUEST);
+    }
+}
+
+bitflags::bitflags! {
+    /// virtio-rtc specific features
+    #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+    struct Feature: u64 {
+        /// Alarm functionality
+        const ALARM                 = 1 << 0;
+
+        // device independent
+        /// Legacy/Unused.
+        const NOTIFY_ON_EMPTY       = 1 << 24; // legacy
+        /// Legacy/Unused.
+        const ANY_LAYOUT            = 1 << 27; // legacy
+        /// Indirect descriptors
+        const RING_INDIRECT_DESC    = 1 << 28;
+        /// `avail_event` and `used_event` fields
+        const RING_EVENT_IDX        = 1 << 29;
+        /// Legacy/Unused.
+        const UNUSED                = 1 << 30; // legacy
+        /// VirtIO version 1 compliance
+        const VERSION_1             = 1 << 32; // detect legacy
+
+        // since virtio v1.1
+        /// Limited device access to memory
+        const ACCESS_PLATFORM       = 1 << 33;
+        /// Packed virtqueue layout.
+        const RING_PACKED           = 1 << 34;
+        /// Optimisations for in-order buffer usage
+        const IN_ORDER              = 1 << 35;
+        /// Platform ordering for memory access
+        const ORDER_PLATFORM        = 1 << 36;
+        /// Single root I/O virtualization
+        const SR_IOV                = 1 << 37;
+        /// Extra data in device notifications
+        const NOTIFICATION_DATA     = 1 << 38;
+
+        // since virtio v1.2
+        /// Driver uses the data provided by the device as a virtqueue identifier in available
+        /// buffer notifications.
+        const NOTIF_CONFIG_DATA     = 1 << 39;
+        /// Driver can reset a queue individually
+        const RING_RESET            = 1 << 40;
+
+        // since virtio v1.3
+        /// Device exposes one or more administration virtqueues.
+        const ADMIN_VQ              = 1 << 41;
+
+        // since virtio v1.4
+        /// the driver can suspend the device by set the `SUSPEND` bit to 1.
+        const SUSPEND               = 1 << 43;
+    }
+}
+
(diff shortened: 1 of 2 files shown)


src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+use crate::transport::Transport;
+use crate::{Error, PAGE_SIZE, Result};
+use alloc::boxed::Box;
+use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes, KnownLayout};
+
+const QUEUE_REQUEST: u16 = 0;
+const QUEUE_ALARM: u16 = 1;
+const SUPPORTED_FEATURES: Feature = Feature::RING_EVENT_IDX
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::ALARM);
```
qwandor: This should probably include `VERSION_1` and `ACCESS_PLATFORM` too.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+
+// a parameter that can change
```
qwandor: That's not a very meaningful comment.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+                negotiated_features.contains(Feature::RING_INDIRECT_DESC),
+                negotiated_features.contains(Feature::RING_EVENT_IDX),
+                negotiated_features.contains(Feature::ACCESS_PLATFORM),
+            )?)
+        } else {
+            None
+        };
+        if request_queue.should_notify() {
+            transport.notify(QUEUE_REQUEST);
+        }
```
qwandor: Why do you need to notify the queue here, when you haven't added any buffers to it yet?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+            alarm_queue,
+            queue_buf_send,
+            queue_buf_recv,
+        })
+    }
+
+    /// Return any unrecognized feature flag bits.
+    ///
+    /// Any bits that cannot be converted to [`Feature`] will be returned.
+    pub fn invalid_feature_bits(&mut self) -> Option<core::num::NonZeroU64> {
```
qwandor: How do you expect this method to be used?

Also, add a use statement for `NonZeroU64` rather than using the full path here.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+        let device_feature_bits = self.transport.read_device_features();
+        let device_features: Feature = Feature::from_bits_truncate(device_feature_bits);
+        let invalid_features = device_feature_bits & (!device_features.bits());
```
qwandor: Can't you just call `device_feature_bits.unknown_bits()`?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+            ..VirtioRtcReqClockCap::default()
+        };
+        let resp: VirtioRtcRespClockCap = self.request(req)?;
+        let kind = match resp.type_ {
+            self::VIRTIO_RTC_CLOCK_UTC => ClockType::Utc,
+            self::VIRTIO_RTC_CLOCK_TAI => ClockType::Tai,
+            self::VIRTIO_RTC_CLOCK_MONOTONIC => ClockType::Monotonic,
+            self::VIRTIO_RTC_CLOCK_UTC_SMEARED => ClockType::UtcSmeared,
+            self::VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED => ClockType::UtcMaybeSmeared,
+            _ => panic!(),
```
qwandor: Return an error rather than panicking, a misbehaving device shouldn't result in a panic.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+            head: VirtioRtcReqHead {
+                msg_type: VIRTIO_RTC_REQ_CLOCK_CAP,
+                ..VirtioRtcReqHead::default()
+            },
+            clock_id,
+            ..VirtioRtcReqClockCap::default()
+        };
+        let resp: VirtioRtcRespClockCap = self.request(req)?;
+        let kind = match resp.type_ {
+            self::VIRTIO_RTC_CLOCK_UTC => ClockType::Utc,
```
qwandor: Why do you need the `self::` for these lines?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+            self::VIRTIO_RTC_CLOCK_TAI => ClockType::Tai,
+            self::VIRTIO_RTC_CLOCK_MONOTONIC => ClockType::Monotonic,
+            self::VIRTIO_RTC_CLOCK_UTC_SMEARED => ClockType::UtcSmeared,
+            self::VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED => ClockType::UtcMaybeSmeared,
+            _ => panic!(),
+        };
+        Ok(ClockCapabilities {
+            kind,
+            leap_second_smearing: None,
```
qwandor: If this is always `None`, what's the point of having the field?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+            alarm_capability: resp.flags > 0,
```
qwandor: What happens if other flags are added in future, shouldn't this check a specific bit?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+unsafe impl<H: Hal, T: Transport + Send> Send for VirtIORTC<H, T> where
+    VirtQueue<H, QUEUE_SIZE>: Send
+{
+}
+
+// SAFETY: An '&VirtIORTC` can't do anything, all methods take `&mut self`.
+unsafe impl<H: Hal, T: Transport + Sync> Sync for VirtIORTC<H, T> where
+    VirtQueue<H, QUEUE_SIZE>: Sync
+{
+}
```
qwandor: Why isn't `VirtIORTC` already `Send` and `Sync` automatically?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::ALARM);
+
+// a parameter that can change
+const QUEUE_SIZE: usize = 64;
+
+/// Driver for RTC devices based on VIRTIO 1.4.
+///
+/// Supports alarm feature.
+pub struct VirtIORTC<H: Hal, T: Transport> {
```
qwandor: For consistency with other driver names, let's call this `VirtIORtc`.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+
+    /// Helper method to send a request to the device and block for a response.
+    fn request<Req: IntoBytes + Immutable, Rsp: FromBytes>(&mut self, req: Req) -> Result<Rsp> {
+        req.write_to_prefix(&mut self.queue_buf_send).unwrap();
+        self.request_queue.add_notify_wait_pop(
+            &[&self.queue_buf_send],
+            &mut [&mut self.queue_buf_recv],
+            &mut self.transport,
+        )?;
```
qwandor, line 126: What's the point of the extra copy to `queue_buf_send` here? Why not just pass `req.as_bytes()` directly to `add_notify_wait_pop`?

Likewise if `Rsp` also implements `IntoBytes` then you can allocate it on the stack directly and then pass `response.as_mut_bytes()` directly for the response.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+                negotiated_features.contains(Feature::ACCESS_PLATFORM),
+            )?)
+        } else {
+            None
+        };
+        if request_queue.should_notify() {
+            transport.notify(QUEUE_REQUEST);
+        }
+        let queue_buf_send = FromZeros::new_box_zeroed_with_elems(PAGE_SIZE).unwrap();
+        let queue_buf_recv = FromZeros::new_box_zeroed_with_elems(PAGE_SIZE).unwrap();
```
qwandor: Do these really need to be 4 KiB long? But see my other comments below, I think you can avoid heap allocation entirely.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+    #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+    pub struct Feature: u64 {
+        /// Alarm functionality
+        #[doc(alias="VIRTIO_RTC_F_ALARM")]
+        const ALARM                 = 1 << 0;
+
+        // device independent
+        ///
+        const NOTIFY_ON_EMPTY       = 1 << 24; // legacy
+        ///
```
qwandor: Why all the empty doc comments?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+        // after they have been freed.
+        self.transport.queue_unset(QUEUE_REQUEST);
+        self.transport.queue_unset(QUEUE_ALARM);
+    }
+}
+
+bitflags::bitflags! {
+    /// virtio-rtc specific features
+    #[derive(Copy, Clone, Debug, Default, Eq, PartialEq)]
+    pub struct Feature: u64 {
```
qwandor: Why does this need to be public?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+pub const VIRTIO_RTC_COUNTER_ARM_VCT: u8 = 0;
+/// x86 Time-Stamp Counter
+pub const VIRTIO_RTC_COUNTER_X86_TSC: u8 = 1;
+/// Invalid
+pub const VIRTIO_RTC_COUNTER_INVALID: u8 = 0xFF;
+
+// Control Requests
+
+/// Discovers the number of clocks
+pub const VIRTIO_RTC_REQ_CFG: u16 = 0x1000;
```
qwandor: Why does this need to be public?

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+    /// second (similar to [`ClockType::UtcSmeared`]), or steps at the start of leap seconds like
+    /// [`ClockType::Utc`].
+    ///
+    /// A clock of type [`ClockType::UtcMaybeSmeared`] can change this behavior for every leap second.
+    UtcMaybeSmeared = VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED,
+}
+
+#[derive(Copy, Clone, Debug)]
+#[repr(u8)]
+/// Leap second smearing variants describe the deviation from the UTC standard in the vicinity of a leap second.
```
qwandor, line 317: Nit: move the doc comment before the attributes.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+struct VirtioRtcRespClockCap {
+    head: VirtioRtcRespHead,
+    type_: u8,
+    leap_second_smearing: u8,
+    flags: u8,
+    reserved: [u8; 5],
+}
+
+/// Clock source variant.
+#[derive(Copy, Clone, Debug)]
```
qwandor: Deriving `Eq` and `PartialEq` may also be worthwhile.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+    UtcSmeared = VIRTIO_RTC_CLOCK_UTC_SMEARED,
+    /// This clock either deviates from the UTC standard by smearing time in the vicinity of a leap
+    /// second (similar to [`ClockType::UtcSmeared`]), or steps at the start of leap seconds like
+    /// [`ClockType::Utc`].
+    ///
+    /// A clock of type [`ClockType::UtcMaybeSmeared`] can change this behavior for every leap second.
+    UtcMaybeSmeared = VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED,
+}
+
+#[derive(Copy, Clone, Debug)]
```
qwandor: Deriving `Eq` and `PartialEq` may also be worthwhile.

src/device/rtc.rs
```diff
@@ -0,0 +1,445 @@
...
+/// Leap second smearing variants describe the deviation from the UTC standard in the vicinity of a leap second.
+pub enum SmearingVariant {
+    /// Specifies a linear smear from noon prior to the leap second until noon after the leap second.
+    NoonLinear = VIRTIO_RTC_SMEAR_NOON_LINEAR,
+    /// Specifies a linear smear as per the [`UTC-SLS`](https://www.cl.cam.ac.uk/~mgk25/time/utc-sls/) proposal.
+    UtcSls = VIRTIO_RTC_SMEAR_UTC_SLS,
+}
+
+/// The capabilities of a specific clock.
+#[derive(Copy, Clone, Debug)]
```
qwandor: Deriving `Eq` and `PartialEq` may also be worthwhile.

src/transport/mod.rs
```diff
@@ -223,7 +223,7 @@ pub enum DeviceType {
     VirtioCAIF = 12,
     MemoryBalloon = 13,
     GPU = 16,
-    Timer = 17,
+    RTC = 17,
```
qwandor: Renaming this enum variant is a breaking change, which I'd rather avoid for now. How about just adding a doc comment explaining that it is usually called RTC instead?

epilys: It was changed in 1.4 (Section `5 Device Types`). I think adding an associated constant as an alias would be backwards compatible. Otherwise we can keep the `Timer` name. WDYT?

src/device/rtc.rs
epilys: You cannot construct a bitflags value with unknown bits, unless you define the unknown bits in an associated constant inside the `bitflags!` macro https://docs.rs/bitflags/latest/bitflags/index.html#externally-defined-flags

src/device/rtc.rs
epilys: We use this to test virtio device implementations in our virtio test suite (unpublished for the moment being)

epilys:
Also removed the alarm feature (and consequently dependence on `alloc`), since support was incomplete

src/device/rtc.rs
qwandor: Doesn't `from_bits_retain` keep the unknown bits?

epilys: Ah, I didn't see that method even existed, thanks.

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+        Ok(Self {
+            transport,
+            negotiated_features,
+            request_queue,
+        })
+    }
+
+    /// Return any unrecognized feature flag bits.
+    ///
+    /// Any bits that cannot be converted to [`Feature`] will be returned.
```
qwandor: Fix the warning here, by removing the link to `Feature`.

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+    pub fn invalid_feature_bits(&mut self) -> Option<NonZeroU64> {
+        let device_feature_bits = self.transport.read_device_features();
+        let device_features: Feature = Feature::from_bits_retain(device_feature_bits);
+        NonZeroU64::new(device_features.unknown_bits())
+    }
+
+    /// Discovers the number of clocks from configuration space.
```
qwandor: This comment seems wrong, it's not using configuration space to check this.

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+            VIRTIO_RTC_CLOCK_MONOTONIC => ClockType::Monotonic,
+            VIRTIO_RTC_CLOCK_UTC_SMEARED => ClockType::UtcSmeared,
+            VIRTIO_RTC_CLOCK_UTC_MAYBE_SMEARED => ClockType::UtcMaybeSmeared,
+            _ => return Err(Error::Unsupported),
+        };
+        let leap_second_smearing = if kind == ClockType::UtcSmeared {
+            match resp.leap_second_smearing {
+                VIRTIO_RTC_SMEAR_UTC_SLS => Some(SmearingVariant::UtcSls),
+                VIRTIO_RTC_SMEAR_NOON_LINEAR => Some(SmearingVariant::NoonLinear),
+                _ => None,
```
qwandor: Do other values actually mean no smearing, or should this be an error instead?

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+            }
+        } else {
+            None
+        };
+        Ok(ClockCapabilities {
+            kind,
+            leap_second_smearing,
+            alarm_capability: resp.flags & VIRTIO_RTC_FLAG_ALARM_CAP > 0,
```
qwandor: Nit: `!= 0` would be more appropriate than `> 0` I think.

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+        let resp: VirtioRtcRespRead = self.request(req)?;
+        Ok(resp.clock_reading)
+    }
+
+    /// Helper method to send a request to the device and block for a response.
+    fn request<Req: IntoBytes + Immutable, Rsp: FromBytes + Default + Immutable + IntoBytes>(
+        &mut self,
+        req: Req,
+    ) -> Result<Rsp> {
+        let mut resp = Rsp::default();
```
qwandor: Optional: you could instead use `FromZeros::new_zeroed()` and avoid requiring `Default`.

src/device/rtc.rs
```diff
@@ -0,0 +1,420 @@
...
+const VIRTIO_RTC_SMEAR_UTC_SLS: u8 = 2;
+
+// Hardware counters
+
+/// Arm Generic Timer Counter-timer Virtual Count Register (`CNTVCT_EL0`)
+const VIRTIO_RTC_COUNTER_ARM_VCT: u8 = 0;
+/// x86 Time-Stamp Counter
+const VIRTIO_RTC_COUNTER_X86_TSC: u8 = 1;
+/// Invalid
+const VIRTIO_RTC_COUNTER_INVALID: u8 = 0xFF;
```
qwandor: Where are these constants used?

src/transport/mod.rs
qwandor: Let's keep the name as it is for now. You can leave a TODO to change it next time we make other breaking changes if you want.

epilys: reverted

src/device/rtc.rs
```diff
@@ -0,0 +1,412 @@
...
+const SUPPORTED_FEATURES: Feature = Feature::RING_EVENT_IDX
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::VERSION_1)
+    .union(Feature::ACCESS_PLATFORM);
+
+/// Driver for RTC devices based on VIRTIO 1.4.
+pub struct VirtIORtc<H: Hal, T: Transport> {
+    transport: T,
+    negotiated_features: Feature,
+    request_queue: VirtQueue<H, 64>,
```
qwandor: How was this queue size chosen? Given that you only ever have one request in flight it should be able to be significantly smaller.

epilys: It's been a year since I wrote this so I don't recall, when I have time I will try to experiment with smaller values and get back to you.

qwandor: I'm pretty sure 4 or 8 should be enough.

qwandor (approved):

PR #249, opened 2026-05-21
perf: optimize virtio-blk queue size and memory barrier

cg24-THU:
Two performance improvements for the virtio-blk driver.

1. Queue size 16 to 256: matches Linux virtio_blk default, reduces per-request notification overhead.
2. Memory barrier SeqCst to Release: matches Linux dma_wmb semantics. On RISC-V changes fence rw,rw to fence rw,w.

Benchmark on StarryOS riscv64 QEMU TCG 10MB sequential IO:
READ 4K: 35.74 to 42.09 MB/s (+17.8%)
WRITE 4K: 1.11 to 1.15 MB/s (+3.6%)

Test plan: cargo check passes, StarryOS smoke test no regression.

### DIFF
--- a/src/device/blk.rs
+++ b/src/device/blk.rs
@@ -10,7 +10,7 @@ use log::info;
 use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
 
 const QUEUE: u16 = 0;
-const QUEUE_SIZE: u16 = 16;
+const QUEUE_SIZE: u16 = 256;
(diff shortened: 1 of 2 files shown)


qwandor:
Are you using `write_blocks_nb` / `read_blocks_nb`? And do you know how much of the performance increase comes from the queue size change versus the memory barrier change? Making the queue 16 times bigger seems like a big change.

PR #251, opened 2026-06-15
Add balloon driver

mengzhuo:
Supports inflate, deflate, memory statistics, free page hinting, and page reporting.

### DIFF
--- a/src/device/balloon.rs
+++ b/src/device/balloon.rs
@@ -1,0 +1,913 @@
+//! Driver for VirtIO memory balloon devices.
+//!
+//! The balloon device allows a guest to give memory back to the host (inflate)
+//! and reclaim it later (deflate). It also supports reporting memory statistics,
+//! free page hinting for live migration, and unused page reporting.
+
+use crate::Result;
+use crate::config::{ReadOnly, ReadWrite, read_config, write_config};
+use crate::hal::Hal;
+use crate::queue::VirtQueue;
+use crate::transport::{InterruptStatus, Transport};
+use log::info;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+const INFLATEQ: u16 = 0;
+const DEFLATEQ: u16 = 1;
+const STATSQ: u16 = 2;
+const FREE_PAGEQ: u16 = 3;
+const REPORTINGQ: u16 = 4;
+const QUEUE_SIZE: u16 = 64;
+
+/// Command ID indicating the device wants the driver to stop hinting.
+pub const CMD_ID_STOP: u32 = 0;
+/// Command ID the driver writes when all free pages have been reported.
+pub const CMD_ID_DONE: u32 = 1;
+
+const SUPPORTED_FEATURES: Feature = Feature::MUST_TELL_HOST
+    .union(Feature::STATS_VQ)
+    .union(Feature::DEFLATE_ON_OOM)
+    .union(Feature::FREE_PAGE_HINT)
+    .union(Feature::PAGE_POISON)
+    .union(Feature::REPORTING)
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::RING_EVENT_IDX)
+    .union(Feature::VERSION_1)
+    .union(Feature::ACCESS_PLATFORM);
+
+/// Driver for a VirtIO memory balloon device.
+///
+/// # Example
+///
+/// ```
+/// # use virtio_drivers::{Error, Hal};
+/// # use virtio_drivers::transport::Transport;
+/// use virtio_drivers::device::balloon::VirtIOBalloon;
+///
+/// # fn example<HalImpl: Hal, T: Transport>(transport: T) -> Result<(), Error> {
+/// let mut balloon = VirtIOBalloon::<HalImpl, _>::new(transport)?;
+///
+/// // Check how many pages the host wants.
+/// let target = balloon.num_pages()?;
+///
+/// // Give 4 pages to the host (inflate).
+/// let pfns = [0x10000_u32, 0x10001, 0x10002, 0x10003];
+/// balloon.inflate(&pfns)?;
+/// balloon.update_actual(4)?;
+///
+/// // Reclaim 2 pages from the host (deflate).
+/// balloon.deflate(&pfns[0..2])?;
+/// balloon.update_actual(2)?;
+///
+/// // Report memory statistics.
+/// if balloon.has_stats() {
+///     let stats = [
+///         virtio_drivers::device::balloon::BalloonStat::new(4, 1024 * 1024),
+///         virtio_drivers::device::balloon::BalloonStat::new(5, 256 * 1024 * 1024),
+///     ];
+///     balloon.report_stats(&stats)?;
+/// }
+/// # Ok(())
+/// # }
+/// ```
+pub struct VirtIOBalloon<H: Hal, T: Transport> {
+    transport: T,
+    negotiated_features: Feature,
+    inflate_queue: VirtQueue<H, { QUEUE_SIZE as usize }>,
+    deflate_queue: VirtQueue<H, { QUEUE_SIZE as usize }>,
+    stats_queue: Option<VirtQueue<H, { QUEUE_SIZE as usize }>>,
+    free_page_queue: Option<VirtQueue<H, { QUEUE_SIZE as usize }>>,
+    reporting_queue: Option<VirtQueue<H, { QUEUE_SIZE as usize }>>,
+}
+
+impl<H: Hal, T: Transport> VirtIOBalloon<H, T> {
+    /// Create a new VirtIO balloon driver.
+    pub fn new(mut transport: T) -> Result<Self> {
+        let negotiated_features = transport.begin_init(SUPPORTED_FEATURES);
+
+        let actual = read_config!(transport, BalloonConfig, actual)?;
+        let num_pages = read_config!(transport, BalloonConfig, num_pages)?;
+        info!(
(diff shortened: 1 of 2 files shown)


src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// These pages must not be accessed by the guest until the host returns
+    /// them via deflation.
+    ///
+    /// Blocks until the device acknowledges the pages.
+    pub fn inflate(&mut self, pfns: &[u32]) -> Result {
+        assert!(!pfns.is_empty(), "Must provide at least one PFN to inflate");
+        assert!(
+            core::mem::size_of_val(pfns) <= QUEUE_SIZE as usize * crate::PAGE_SIZE,
+            "Too many PFNs for the virtqueue"
+        );
```
qwandor, line 188: Why is this required? What's wrong with adding a larger slice to the queue?

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+use crate::transport::{InterruptStatus, Transport};
+use log::info;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+const INFLATEQ: u16 = 0;
+const DEFLATEQ: u16 = 1;
+const STATSQ: u16 = 2;
+const FREE_PAGEQ: u16 = 3;
+const REPORTINGQ: u16 = 4;
+const QUEUE_SIZE: u16 = 64;
```
qwandor, line 20: Why do the queues need to be so big? As far as I can see you only ever add one element to each queue (via `add_notify_wait_pop`) with a small number of buffers, so a queue size of 4 or so should be enough.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+
+    /// Gives pages to the host (inflates the balloon).
+    ///
+    /// `pfns` is a slice of page frame numbers (`physical_address >> 12`).
+    /// These pages must not be accessed by the guest until the host returns
+    /// them via deflation.
+    ///
+    /// Blocks until the device acknowledges the pages.
+    pub fn inflate(&mut self, pfns: &[u32]) -> Result {
+        assert!(!pfns.is_empty(), "Must provide at least one PFN to inflate");
```
qwandor, line 184: Would it make sense for this to be a no-op rather than a panic?

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// Reclaims pages from the host (deflates the balloon).
+    ///
+    /// `pfns` is a slice of page frame numbers (`physical_address >> 12`)
+    /// for pages previously given to the host via `inflate`.
+    ///
+    /// Blocks until the device acknowledges. If `VIRTIO_BALLOON_F_MUST_TELL_HOST`
+    /// was negotiated, the driver must wait for this acknowledgment before
+    /// reusing the pages.
+    pub fn deflate(&mut self, pfns: &[u32]) -> Result {
+        assert!(!pfns.is_empty(), "Must provide at least one PFN to deflate");
```
qwandor, line 205: This also seems like it could just be a no-op.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// Sends free page data on the free page hint virtqueue.
+    ///
+    /// The `data` buffer should contain the memory contents of free pages.
+    /// Blocks until the device acknowledges.
+    pub fn free_page_send_pages(&mut self, data: &[u8]) -> Result {
+        let queue = self
+            .free_page_queue
+            .as_mut()
+            .ok_or(crate::Error::Unsupported)?;
+        queue.add_notify_wait_pop(&[data], &mut [], &mut self.transport)?;
```
qwandor, line 285: Does this work? From what I can tell from reading the specification, free pages should be passed as output buffers rather than input buffers in our terms. (Though the specification seems to name inputs and outputs the opposite way around to us.)

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+            .ok_or(crate::Error::Unsupported)?;
+        queue.add_notify_wait_pop(&[cmd_id.as_bytes()], &mut [], &mut self.transport)?;
+        Ok(())
+    }
+
+    /// Sends free page data on the free page hint virtqueue.
+    ///
+    /// The `data` buffer should contain the memory contents of free pages.
+    /// Blocks until the device acknowledges.
+    pub fn free_page_send_pages(&mut self, data: &[u8]) -> Result {
```
qwandor, line 280: Isn't there a requirement that the guest doesn't use this free page until the device releases the hinted free pages? If so, then this method should be unsafe, and should probably take a list of pages rather than just one at a time. Perhaps the pages should be passed as raw `NonNull` pointers rather than slices?

How does this interact with `Hal::share`, in case the guest has protected memory or an IOMMU or something?

You could consider adding a method to manage the whole process, sending the command ID, then all the free pages, then `CMD_ID_STOP`.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+        )?;
+        let deflate_queue = VirtQueue::new(
+            &mut transport,
+            DEFLATEQ,
+            indirect,
+            event_idx,
+            access_platform,
+        )?;
+
+        let stats_queue = if negotiated_features.contains(Feature::STATS_VQ) {
```
qwandor, line 119: The specification says "After initializing the device, the driver MUST make an output buffer available in the statsq.". I guess that means you need to add a buffer to it immediately here.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// memory state (free memory, swap usage, page faults, etc.).
+    ///
+    /// Blocks until the device acknowledges receipt.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`Error::Unsupported`] if the stats queue was not negotiated.
+    pub fn report_stats(&mut self, stats: &[BalloonStat]) -> Result {
+        let queue = self.stats_queue.as_mut().ok_or(crate::Error::Unsupported)?;
+        queue.add_notify_wait_pop(&[stats.as_bytes()], &mut [], &mut self.transport)?;
```
qwandor, line 315: Is the device expected to use the buffer immediately? The specification makes it sound like not, so this could block for a long time.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// virtqueue.
+    ///
+    /// The `data` buffer should contain page data the host may reclaim.
+    /// Blocks until the device acknowledges.
+    pub fn report_unused_pages(&mut self, data: &[u8]) -> Result {
+        let queue = self
+            .reporting_queue
+            .as_mut()
+            .ok_or(crate::Error::Unsupported)?;
+        queue.add_notify_wait_pop(&[data], &mut [], &mut self.transport)?;
```
qwandor, line 299: Again, does this actually work? It thought from the specification that the free pages should be listed as outputs rather than inputs. And how does it interact with `Hal::share`?

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+        Ok(())
+    }
+
+    /// Reports unused pages to the host for reclamation via the page reporting
+    /// virtqueue.
+    ///
+    /// The `data` buffer should contain page data the host may reclaim.
+    /// Blocks until the device acknowledges.
+    pub fn report_unused_pages(&mut self, data: &[u8]) -> Result {
```
qwandor, line 294: The specification says "The driver MUST NOT use the reported pages until the device has acknowledged the reporting request.", so this should at least take the page as `&mut` rather than `&` to prevent concurrent reads.

However it also says "If the VIRTIO_BALLOON_F_PAGE_POISON feature has not been negotiated, then the driver MUST treat all reported pages as uninitialized memory.". It's not clear to me whether we need to treat it as uninitialised memory from the Rust memory model point of view, or just as arbitrary bytes.

Further, "If the VIRTIO_BALLOON_F_PAGE_POISON feature has been negotiated, the driver MUST initialize all free pages with poison_val before reporting them.", so we need to handle that case.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    }
+
+    /// Gives pages to the host (inflates the balloon).
+    ///
+    /// `pfns` is a slice of page frame numbers (`physical_address >> 12`).
+    /// These pages must not be accessed by the guest until the host returns
+    /// them via deflation.
+    ///
+    /// Blocks until the device acknowledges the pages.
+    pub fn inflate(&mut self, pfns: &[u32]) -> Result {
```
qwandor, line 183: The specification says that guest must not use these pages until the device has acknowledged the deflate request, and if `VIRTIO_BALLOON_F_PAGE_POISON` is negotiated then "The driver MUST initialize the deflated pages with poison_val when they are reused by the driver.". This method should be unsafe with these as safety requirements.

src/device/balloon.rs
```diff
@@ -0,0 +1,913 @@
...
+    ///
+    /// Blocks until the device acknowledges. If `VIRTIO_BALLOON_F_MUST_TELL_HOST`
+    /// was negotiated, the driver must wait for this acknowledgment before
+    /// reusing the pages.
+    pub fn deflate(&mut self, pfns: &[u32]) -> Result {
+        assert!(!pfns.is_empty(), "Must provide at least one PFN to deflate");
+
+        self.deflate_queue
+            .add_notify_wait_pop(&[pfns.as_bytes()], &mut [], &mut self.transport)?;
+
```
qwandor, line 209: Would it make sense for this to also handle initialising the pages to the poison value if `VIRTIO_BALLOON_F_PAGE_POISON` is negotiated? Or is that not practical because the `pfns` are physical addresses rather than virtual addresses?

PR #252, opened 2026-06-15
Add iommu driver

mengzhuo:
Supports ATTACH, DETACH, MAP, UNMAP, and PROBE requests on the request virtqueue, plus fault reporting on the event virtqueue.

### DIFF
--- a/src/device/iommu.rs
+++ b/src/device/iommu.rs
@@ -1,0 +1,873 @@
+//! Driver for VirtIO IOMMU devices.
+//!
+//! The IOMMU device manages DMA remapping for endpoints. It translates I/O
+//! virtual addresses to physical addresses using a request/response protocol
+//! on the request virtqueue, and reports translation faults on the event
+//! virtqueue.
+
+use crate::config::{ReadOnly, read_config};
+use crate::hal::Hal;
+use crate::queue::VirtQueue;
+use crate::transport::Transport;
+use crate::{Error, Result};
+use log::info;
+use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes, KnownLayout};
+
+const QUEUE_REQUEST: u16 = 0;
+const QUEUE_EVENT: u16 = 1;
+const QUEUE_SIZE: u16 = 64;
+
+const SUPPORTED_FEATURES: Feature = Feature::INPUT_RANGE
+    .union(Feature::DOMAIN_RANGE)
+    .union(Feature::MAP_UNMAP)
+    .union(Feature::PROBE)
+    .union(Feature::RING_INDIRECT_DESC)
+    .union(Feature::RING_EVENT_IDX)
+    .union(Feature::VERSION_1)
+    .union(Feature::ACCESS_PLATFORM);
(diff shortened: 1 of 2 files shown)


src/device/iommu.rs
```diff
@@ -0,0 +1,873 @@
...
+    pub fn probe(&mut self, endpoint: u32, buf: &mut [u8]) -> Result<usize> {
+        let len = buf.len();
+        buf[..4].copy_from_slice(
+            VirtioIommuReqHead {
+                type_: VIRTIO_IOMMU_T_PROBE,
+                reserved: Default::default(),
+            }
+            .as_bytes(),
+        );
+        buf[4..8].copy_from_slice(&endpoint.to_le_bytes());
```
qwandor, line 291: Why does this need to go in `buf` rather than either a separate 8-byte long buffer on the stack, or or just passing the request struct and endpoint directly as two elements in the inputs of `add_notify_wait_pop`?

src/device/iommu.rs
```diff
@@ -0,0 +1,873 @@
...
+
+        let (input, output) = buf.split_at_mut(8);
+        self.request_queue
+            .add_notify_wait_pop(&[input], &mut [output], &mut self.transport)?;
```
qwandor, line 295: Aren't there supposed to be 64 reserved bytes between the endpoint and the properties array?

Also, rather than using `buf` for both input and output, I suggest passing the different parts separately, like `add_notify_wait_pop(&[head.as_bytes(), &endpoint.to_le_bytes(), &reserved], &mut [buf, tail.as_bytes_mut()])`. Then you don't have to copy bytes around or split things up.

src/device/iommu.rs
```diff
@@ -0,0 +1,873 @@
...
+
+        let tail_offset = buf.len() - 4;
+        let status = buf[tail_offset];
+        iommu_status_to_result(status)?;
+        Ok(len)
+    }
+
+    /// Returns true if there is a fault event available on the event queue.
+    pub fn fault_available(&self) -> bool {
+        self.event_queue.can_pop()
```
qwandor, line 305: You never add buffers to the event queue, so this will never return true.

src/device/iommu.rs
```diff
@@ -0,0 +1,873 @@
...
+    }
+
+    /// Pops a fault event from the event queue, if one is available.
+    pub fn pop_fault(&mut self) -> Result<Option<VirtioIommuFault>> {
+        if !self.event_queue.can_pop() {
+            return Ok(None);
+        }
+        let mut fault = VirtioIommuFault::new_zeroed();
+        let token = self.event_queue.peek_used().unwrap();
+        // SAFETY: fault buffer is valid for the duration of pop_used.
```
qwandor, line 315: The buffers passed to `pop_used` must be exactly the ones passed to the corresponding earlier call to `add`. You never added this `fault` buffer to the event queue, so this is not sound.

If you don't mind requiring `alloc` then the simplest solution would be to use `OwningQueue` for the `event_queue`.

PR #255, opened 2026-07-06
Feature-gate device and transport drivers

osteffenrh:
Add cargo features to gate individual device and transport drivers.
This is useful for consuming projects, to select only the parts required.

### DIFF
--- a/src/transport/some.rs
+++ b/src/transport/some.rs
@@ -1,37 +1,55 @@
 use zerocopy::{FromBytes, Immutable, IntoBytes};
 
-use super::{DeviceStatus, DeviceType, Transport, mmio::MmioTransport, pci::PciTransport};
+use super::{DeviceStatus, DeviceType, Transport};
+
+#[cfg(feature = "mmio")]
+use super::mmio::MmioTransport;
+#[cfg(feature = "pci")]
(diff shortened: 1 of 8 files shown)


src/transport/some.rs
```diff
@@ -83,7 +101,15 @@ impl Transport for SomeTransport<'_> {
...
+        dispatch!(
+            self,
+            queue_set,
+            queue,
+            size,
+            descriptors,
+            driver_area,
+            device_area
+        )
     }
```
epilys, line 113: unrelated change?

qwandor:
That's a lot of features, and makes the code more complex to follow. What's the benefit of this, a slight improvement in compile time? Unused drivers shouldn't have any impact on code size anyway, unreachable code will be optimised out of the final binary.

src/transport/some.rs
osteffenrh: Oh, strange. Maybe I forgot to run the formatter the first time around.

osteffenrh:
I agree that this adds a lot of features and that the effective gain is small.

The idea came from a build error that we encountered in an unused driver, but that was
solved in the meantime (toolchain update).

Let's close this then.

PR #258, opened 2026-08-14
Remove queue_buf_recv from VirtIOGpu.

qwandor:
Allocate values as needed instead. This saves extra copies.


zyc107109102:
Thanks for the follow-up — I've gone through the changes and they look
great. Dropping the shared `queue_buf_recv` and writing the response
straight into a zeroed struct is cleaner than what I had: it saves a copy
and removes the shared-buffer state I had to work around in `get_capset`.

I've also verified it works on a real OS kernel (tgoskit) — the virgl path
runs correctly there. Happy to help with anything else.

PR #263, opened 2026-08-31
Fix display refresh and vblank timing

kroese:
Use EDID timing data instead of hard-coded 60 Hz values and drive the vblank timer from the active display mode.


src/device/gpu/edid.rs
```diff
@@ -92,23 +92,47 @@ impl StandardTiming {
...
 }
 
 impl DetailedTiming {
     /// Parse a detailed timing descriptor from its 18-byte encoding.
     ///
-    /// Returns `None` if the active pixel counts are zero.
+    /// Returns `None` for a display descriptor (zero pixel clock), or if
+    /// the active pixel counts are zero.
     fn parse(bytes: &[u8; DTD_LEN]) -> Option<Self> {
+        let pixel_clock_10khz = u16::from_le_bytes([bytes[0], bytes[1]]) as u32;
```
qwandor: Nit: use `.into()` or `u32::from(...)` for the conversion to make it clear that it can't lose any precision.

src/device/gpu/edid.rs
```diff
@@ -117,8 +141,8 @@ impl DetailedTiming {
 /// Wraps the raw EDID byte blob and provides methods to extract display
 /// information such as preferred resolution and supported standard timings.
 pub struct Edid {
-    pub(super) data: [u8; 1024],
-    pub(super) size: u32,
+    pub data: [u8; 1024],
+    pub size: u32,
```
qwandor: Do these fields really need to be public? They seem like an internal implementation detail.

src/device/gpu/edid.rs
```diff
@@ -145,13 +169,28 @@ impl Edid {
         StandardTiming::parse(bytes)
     }
 
+    /// Get the preferred detailed timing from the EDID data.
+    ///
+    /// Returns (active width, active height, total width, total height, pixel clock in Hz)
+    /// from the first Detailed Timing Descriptor.
+    pub fn preferred_timing(&self) -> Result<(u32, u32, u32, u32, u32)> {
```
qwandor: Returning a long tuple like this is easy for the caller to get wrong. Would it make sense to make `DetailedTiming` public and return that? Or else to make a new public struct with the relevant fields.

qwandor:
> Use EDID timing data instead of hard-coded 60 Hz values and drive the vblank timer from the active display mode.

This doesn't seem to be describing what this change actually does, which is just to make the timing data available. Did you mean to update the examples to use it, perhaps?

kroese:
> > Use EDID timing data instead of hard-coded 60 Hz values and drive the vblank timer from the active display mode.
> 
> This doesn't seem to be describing what this change actually does, which is just to make the timing data available. Did you mean to update the examples to use it, perhaps?

No, I wanted to implement the correct refresh rate in https://github.com/anonymix007/virtio-drivers-windows-rs/commit/8e60e79c890e5d5865c90dfa036ff12f884ed78b and since `edid.rs` came from upstream, I submitted this patch here instead of modifying the copy that repo contains.

src/device/gpu/edid.rs
kroese: Should be resolved now, thanks!

src/device/gpu/edid.rs
kroese: Should be resolved now, thanks!

src/device/gpu/edid.rs
kroese: Should be resolved now, thanks!

kroese:
@qwandor I addressed your feedback

qwandor (approved):

PR #265, opened 2026-09-11
gpu: async fire-and-forget control queue with Linux-style batching

zyc107109102:
# Code Walkthrough: virtio-gpu Async Control Queue (commit `0496c58`)
> Reference: Linux `drivers/gpu/drm/virtio/` (virtgpu_vq.c / virtgpu_submit.c / virtgpu_ioctl.c).

## 0. Big picture: call chain and file layout

```
Consumer (OS / example)
   │
   ├─ Blocking API (mod.rs, upstream method names, unchanged semantics)   get_display_info / resource_create_2d / submit_3d / …
   │      └────────────┐
   ├─ Async API (async_api.rs, 9 *_async + 4 support fns) ───────────────┤
   │                                                                     ▼
   │                                        ControlQueue (ctrl.rs, new file)
   │                                     enqueue / request_sync / pump_completions
   │                                     / wait_fence / notify / flush_pending
   │                                                                     ▼
   └─ Cursor API (mod.rs) ──→ cursor_queue: VirtQueue ──→ VirtQueue (queue.rs, zero diff vs upstream)
                                                                 ▼
                                                        virtqueue ring → QEMU
```

- `mod.rs`: blocking methods keep their upstream shape; all command structs (`ResourceCreate2D` etc.) and `CtrlHeader`.
- `async_api.rs`: all async entry points — thin wrappers that build **exactly the same** command structs as the blocking versions, submitted a different way.
- `ctrl.rs`: the machinery itself — ~120 lines of data structures, ~400 lines of core logic, ~320 lines of tests.

---

### 1.1 `PendingSubmit` (ctrl.rs:68) — the ledger entry for one in-flight async command

```rust
struct PendingSubmit {
    slot: usize,              // index into the vbuf arena
    cmd_len: usize,           // valid bytes in vbuf_cmds[slot] (command sizes vary)
    data: Option<Box<[u8]>>,  // optional second payload (SUBMIT_3D's virgl command stream), heap
    fence_id: u64,            // upper-layer fence id (non-zero for SUBMIT_3D only)
}
```

Note it does **not** contain the command bytes — only "where the bytes are". Reason in 1.2.

### 1.2 `ControlQueue` (ctrl.rs:91) — field by field

```rust
pub(crate) struct ControlQueue<H: Hal> {
    queue: VirtQueue<H, { CTRL_QUEUE_SIZE as usize }>,   // the 64-deep ring itself
    vbuf_cmds:  Box<[[u8; INLINE_CMD_SIZE]]>,  // 96B × 192 slots, command-byte arena
    vbuf_resps: Box<[[u8; RESP_SIZE]]>,        // 24B × 192 slots, response arena, indexed in parallel
    vbuf_free:  Vec<usize>,                    // free-slot stack (no ordering required)
    pending: [Option<PendingSubmit>; 64],      // in-flight commands, indexed directly by descriptor token
    pending_commands: VecDeque<PendingSubmit>, // commands parked while the ring was full (cap 128)
    ctrl_pending: u32,                         // enqueues accumulated since the last kick decision
    completed_fence_id: u64,                   // fence high-water mark: ≥N done ⇒ all ≤N done
}
```

Three key design decisions:

**(a) Why do the command bytes live in `Box<[[u8; 96]]>` instead of inside `PendingSubmit`?**
Fire-and-forget means the command bytes must still be DMA-able by the device at an **arbitrary later time**, after `enqueue` has returned and the caller's stack is gone. A `PendingSubmit` moves around inside the driver (into `pending[token]`, into the parking FIFO); if the 96 bytes were inlined, every move would leave the descriptors pointing at dead memory. So the **bytes live in an arena that never moves, and the struct carries only `slot: usize`** — the index travels, the bytes stay put. This is not theoretical hygiene: during development we caught a real bug where moving the struct made the device DMA garbage commands.

**Mirrors**: Linux's `kmem_cache`-allocated `virtio_gpu_vbuf` (the `vbufs` slab in virtgpu_vq.c) — likewise a pre-allocated, address-stable pool of command buffers. The 96-byte inline cap = Linux `MAX_INLINE_CMD_SIZE` (the largest control command, `CmdCtxCreate`, is exactly 96 bytes); `RESP_SIZE = size_of::<CtrlHeader>() = 24`.

**(b) Why two parallel arenas (`vbuf_cmds` / `vbuf_resps`) instead of one?**
Borrow splitting. Every submission must hand `queue.add()` the command slot (device-readable, shared reference) and the response slot (device-writable, mutable reference) **simultaneously**. In a single arena the two indexes would conflict in the borrow checker's view; as two fields, `entry_bufs` can borrow `&self.vbuf_cmds` and `&mut self.vbuf_resps` at once:

```rust
fn entry_bufs<'a>(entry: &'a PendingSubmit,
                  cmds: &'a [[u8; 96]], resps: &'a mut [[u8; 24]])
                 -> (&'a [u8], Option<&'a [u8]>, &'a mut [u8]) {
    (&cmds[entry.slot][..entry.cmd_len], entry.data.as_deref(), &mut resps[entry.slot])
}
```

**(c) Why is `pending` a 64-entry array indexed directly by token?**
`queue.add()` returns the descriptor-table index as token, always in `0..64` (the free-list head is a descriptor-table index). token → `pending[token]` is O(1) with no lookup structure. Commands parked in the FIFO are **not on the ring and have no token**, which is why `PendingSubmit` has no token field — the token is decided by `add_pending` at re-add time.

### 1.3 Memory layout at a glance

```
 vbuf_cmds: Box<[[u8;96]; 192]>      vbuf_resps: Box<[[u8;24]; 192]>
   slot 0 ┌────────────┐               slot 0 ┌──────────┐
          │ cmd bytes  │  addresses fixed       │ resp 24B │
   slot 191└────────────┘  for life        slot 191└──────────┘
              ▲   ▲                            ▲
              │   │ the slot index travels with the struct; the bytes never move
   pending[0..64]: Option<PendingSubmit> ──────┘
        ▲
        │ token (= descriptor-table index 0..63)
   in-flight commands on the ring
   pending_commands (VecDeque, ≤128): parked, no token yet
```

---

## 2. Life of an async command: `enqueue()` (ctrl.rs:172)

Take `resource_create_2d_async(100, 1280, 800)` → `ctrl.enqueue(&transport, &ResourceCreate2D{..}, None, 0)`:

### Step 1: flush the parking FIFO first (the root of ordering)

```rust
self.flush_pending(transport)?;
```

Parked commands were submitted **earlier**. They must re-enter the ring before this command — virtqueue strict FIFO is the foundation of the whole "no per-command fence needed" argument, and that foundation is maintained by flushing at every submission point (`enqueue`, `add_sync`).

### Step 2: validate and copy the payload

```rust
if cmd_bytes.len() > INLINE_CMD_SIZE { return Err(Error::InvalidParam); }
let mut data: Option<Box<[u8]>> = data.map(<[u8]>::to_vec).map(Vec::into_boxed_slice);
```

Command bodies over 96B are rejected outright (all control commands are ≤96B; larger means the caller built something wrong). The second payload is copied to a heap box immediately — **from this point on it is independent of the caller's stack**.

### Step 3: acquire a slot and hit the ring — the `held` loop

```rust
let mut held: Option<PendingSubmit> = None;
loop {
    let mut entry = match held.take() {
        Some(entry) => entry,
        None => {
            let Some(slot) = self.vbuf_free.pop() else {
                self.pump_completions(transport)?;   // all 192 slots live: reclaim to make room
                spin_loop();
                continue;
            };
            self.vbuf_cmds[slot][..cmd_bytes.len()].copy_from_slice(cmd_bytes);
            self.vbuf_resps[slot] = [0; RESP_SIZE];
            PendingSubmit { slot, cmd_len: .., data: data.take(), fence_id }
        }
    };
    match self.add_pending(&mut entry) {             // hit the ring, do NOT kick!
        Ok(token) => {
            self.account_enqueued(transport);        // count + maybe batch-kick
            self.pending[token as usize] = Some(entry);
            return Ok(Some(token));                  // on the ring
        }
        Err(Error::QueueFull) => {
            self.pump_completions(transport)?;       // reclaim once, retry once
            match self.add_pending(&mut entry) {
                Ok(token) => { /* same as above, return Ok(Some(token)) */ }
                Err(Error::QueueFull) if self.pending_commands.len() < PENDING_FIFO_CAP => {
                    self.pending_commands.push_back(entry);
                    return Ok(None);                 // parked — but still not blocking
                }
                Err(Error::QueueFull) => {
                    held = Some(entry);              // FIFO full too: keep holding the slot
                    spin_loop();                     // wait for the host to drain, retry
                }
                Err(e) => { self.vbuf_free.push(entry.slot); return Err(e); }
            }
        }
        Err(e) => { self.vbuf_free.push(entry.slot); return Err(e); }
    }
}
```

The non-obvious points, one by one:

- **Why `held` exists**: once constructed, a `PendingSubmit` **owns** an arena slot. When the FIFO is also full we can neither drop it (slot leak) nor re-copy the payload (a SUBMIT_3D command stream is hundreds of KB). The fix: stash it in `held` and loop back — the code comment states that each loop iteration either reuses `held` or returns, so the payload box is **moved, never cloned**.
- **`data.take()`**: same reason — the payload box's ownership moves exactly once, from the caller's `&[u8]` → heap box → `PendingSubmit`, with at most one copy end to end.
- **Progress guarantee**: both spin points (vbuf exhaustion, ring+FIFO full) call `pump_completions` every round, and **every reclaimed used entry returns exactly one vbuf slot** (see §5). As long as the host is making progress, the wait must terminate — it cannot degrade into an unbounded spin.
- **Three-state return**: `Ok(Some(token))` = on the ring; `Ok(None)` = parked (swallowed by the `*_async` wrappers, §8); `Err` only for bad parameters / transport errors.

**Mirrors**: enqueue-and-return, never block the submitter on host drain — the `virtio_gpu_queue_ctrl_sgs` semantics. Linux's submitter sleeps on a waitqueue when full; ours doesn't even sleep (it parks), which goes beyond Linux — paid for with a **bounded** 192-slot guest buffer.

### `add_pending()` (ctrl.rs:364) — hit the ring, never kick

```rust
fn add_pending(&mut self, entry: &mut PendingSubmit) -> Result<u16> {
    let (cmd, data, resp) = entry_bufs(entry, &self.vbuf_cmds, &mut self.vbuf_resps);
    match data {
        Some(d) => unsafe { self.queue.add(&[cmd, d], &mut [resp]) },
        None    => unsafe { self.queue.add(&[cmd], &mut [resp]) },
    }
}
```

`queue.add` only writes the descriptor table and the avail-ring index — it **does not notify the device**. The command is now "visible but not necessarily announced" to the host. Whether and when to actually kick is the next function's business.

---

## 3. How commands reach the host: `account_enqueued` → `maybe_kick` (ctrl.rs:380/396)

```rust
fn account_enqueued(&mut self, transport: &mut impl Transport) {
    self.ctrl_pending = self.ctrl_pending.saturating_add(1);
    if self.ctrl_pending >= KICK_THRESHOLD || self.queue.available_desc() <= 2 {
        self.maybe_kick(transport, true);          // threshold reached / ring nearly full: forced
    }
}

fn maybe_kick(&mut self, transport: &mut impl Transport, force: bool) {
    if !force && self.ctrl_pending == 0 {
        return;                                     // empty accumulator: no decision at all
    }
    self.ctrl_pending = 0;                          // zeroed on every DECISION (see below)
    if force || self.queue.available_desc() <= 2 || self.queue.should_notify() {
        transport.notify(QUEUE_TRANSMIT);           // the only MMIO-write exit
    }
}
```

Four points worth unpacking:

1. **`ctrl_pending` is zeroed on every kick *decision*, not only on an actual MMIO write.** If `should_notify()` is false (under event_idx: the host is not waiting for a kick = the host is actively draining), the commands are already on the avail ring and will be processed — accumulating the counter further would be meaningless.
2. **`should_notify()` ≡ upstream `virtqueue_kick_prepare`**: under event_idx it reads the host-written `avail_event`; `avail_idx >= avail_event + 1` means "host asleep, waiting for a kick" → must write MMIO; otherwise (host running) skip it. A burst of commands within one host-drain cycle coalesces into roughly one MMIO write.
3. **Two forced lines**: the threshold of 8 guarantees that a consumer which *never* calls notify still gets every command delivered by the 8th at the latest (a safety property, locked by the `fire_and_forget_commands_are_delivered_without_caller_notify` test); `available_desc() <= 2` is real backpressure — a stalled host must be nudged.
4. **The public `ctrl_notify()` (ctrl.rs:416) = `maybe_kick(force=false)`**: a consumer with a transaction boundary calls it once at the end of a batch, delivering the whole batch immediately (what Linux's `virtio_gpu_notify()` does at the ioctl boundary); with an empty accumulator it is a strict no-op (tested).

**Difference from Linux and why**: Linux defers the kick to the DRM ioctl boundary, relying on the contract that all submissions come from ioctls; a no-std crate cannot assume such a boundary exists. The solution is **bounded deferral**: the threshold of 8 backstops consumers without a boundary so commands can never starve (that safety property has a dedicated test), while boundary-aware consumers get Linux-identical per-transaction kicks via `ctrl_notify()`.

---

## 4. When the ring is full: the parking FIFO and `flush_pending` (ctrl.rs:494)

```rust
fn flush_pending(&mut self, transport: &mut impl Transport) -> Result {
    let mut re_added = false;
    while let Some(mut cmd) = self.pending_commands.pop_front() {  // strict FIFO dequeue
        match self.add_pending(&mut cmd) {
            Ok(token) => {
                self.pending[token as usize] = Some(cmd);
                self.ctrl_pending = self.ctrl_pending.saturating_add(1);  // counts toward the accumulator!
                re_added = true;
            }
            Err(Error::QueueFull) => {
                self.pending_commands.push_front(cmd);  // full again: back to the front, order kept
                break;
            }
            Err(e) => { self.pending_commands.push_front(cmd); return Err(e); }
        }
    }
    if re_added {
        self.maybe_kick(transport, false);   // re-added commands need a kick decision
    }
    Ok(())
}
```

- **Re-added commands count into the accumulator**: if nobody calls notify afterwards, they are covered by the next threshold kick — they can't be forgotten after re-entering the ring.
- **`push_front` rollback**: if the ring saturates mid-flush, the unsuccessful commands go back to the **front**, preserving first-submitted-first-on-ring.
- **Call sites**: the start of `enqueue`/`add_sync` (ordering), and the **end** of `pump_completions` — freed slots are backfilled immediately. So even if the guest stops submitting for a while, IRQ-driven pumps keep draining the parking FIFO.

---

## 5. Harvesting completions: `pump_completions` (ctrl.rs:434)

```rust
pub(crate) fn pump_completions(&mut self, transport: &mut impl Transport) -> Result {
    while self.queue.can_pop() {
        let Some(token) = self.queue.peek_used() else { break };
        let Some(entry) = self.pending[token as usize].take() else { break };   // ← key
        {
            let (cmd, data, resp) = entry_bufs(&entry, &self.vbuf_cmds, &mut self.vbuf_resps);
            let popped = match data {
                Some(d) => unsafe { self.queue.pop_used(token, &[cmd, d], &mut [resp]) },
                None    => unsafe { self.queue.pop_used(token, &[cmd], &mut [resp]) },
            };
            if let Err(e) = popped {
                self.vbuf_free.push(entry.slot);    // pop failed: return the slot before propagating
                return Err(e);
            }
        }
        if entry.fence_id > self.completed_fence_id {
            self.completed_fence_id = entry.fence_id;                        // fence advance
        }
        let rsp_hdr = CtrlHeader::read_from_bytes(&self.vbuf_resps[entry.slot])..;
        if rsp_hdr.hdr_type.0 >= Command::ERR_UNSPEC.0 {
            log::warn!("virtio-gpu: control command 0x{:x} (token {}) failed ...");  // errors visible
        }
        self.vbuf_free.push(entry.slot);            // slot back to the pool
    }
    self.flush_pending(transport)                   // freed slots let parked commands back on
}
```

Four non-obvious mechanisms:

1. **`take()` is a claim.** It removes the `Option` — if a same-core IRQ interrupts this loop and the handler runs pump again, it sees `None` for the same token and breaks; it can never double-pop. The interrupted iteration still holds `entry` locally and completes its own `pop_used` when resumed.
2. **The second meaning of `else { break }`**: `pending[token] == None` can also mean a **synchronous request's entry** — sync entries never get a `pending` record (their bytes live on the caller's stack, §6). The used ring is strictly FIFO; while the head (the sync entry) can't be processed, nothing behind it can be complete either, so breaking is correct and necessary: **an IRQ-path pump can never steal a sync waiter's completion**.
3. **Error logging**: fire-and-forget has no return channel, so device error responses (`≥ ERR_UNSPEC`) are read from the two arenas at reclaim time and surfaced via `log::warn` — matching Linux `virtio_gpu_dequeue_ctrl_func`, which logs error responses too.
4. **The fence advance happens only here (async side)**: `entry.fence_id` advances the monotonic high-water mark; "N done ⇒ all ≤N done" follows from FIFO, so the check is O(1). **Invariant**: every path that pops a fenced entry must advance the counter — the other one is `request_sync_fenced` in §6.

**Mirrors**: Linux `virtio_gpu_dequeue_ctrl_func` (IRQ-driven reclaim). The function is designed to be callable from both the IRQ handler and polling paths — it never sleeps and only reclaims what has already completed.

---

### 6.1 Why upstream's `add_notify_wait_pop` cannot be used directly

It is this driver's ancestor, but it has two structural defects under mixed traffic:
1. It pops blindly, assuming it is the only command on the queue — with fire-and-forget commands queued ahead it pops **someone else's token** and returns a spurious `WrongToken`;
2. On a full ring it propagates `QueueFull` to the caller, with no "kick the host and retry" recovery.

### 6.2 `add_sync` (ctrl.rs:306) — borrowed stack buffers onto the ring + kick

```rust
fn add_sync<'a, 'b>(&mut self, transport: &mut impl Transport,
                    inputs: &'a [&'b [u8]], outputs: &'a mut [&'b mut [u8]]) -> Result<u16> {
    self.flush_pending(transport)?;                  // parked commands first (ordering, as in enqueue)
    let token = loop {
        match unsafe { self.queue.add(inputs, outputs) } {
            Ok(t) => break t,
            Err(Error::QueueFull) => {
                transport.notify(QUEUE_TRANSMIT);    // a response is owed: unconditional kick
                self.pump_completions(transport)?;
                spin_loop();
            }
            Err(e) => return Err(e),
        }
    };
    if self.queue.should_notify() {
        transport.notify(QUEUE_TRANSMIT);            // identical to upstream's kick decision
    }
    Ok(token)
}
```

- The `unsafe` obligation: the borrowed caller-stack buffers must be paired with a `pop_used` **within this frame** — an obligation discharged by `wait_sync`.
- The QueueFull-retry `notify` is **unconditional**: a sync command is owed a response, so the host must be pushed. The normal-path kick goes through `should_notify` suppression — word-for-word the same as the add/kick half of upstream `add_notify_wait_pop`.

### 6.3 `wait_sync` (ctrl.rs:340) — wait for *your own* token only

```rust
loop {
    self.pump_completions(transport)?;               // reclaims earlier async commands for free
    if self.queue.peek_used() == Some(token) {
        return unsafe { self.queue.pop_used(token, inputs, outputs) };  // device writes in place
    }
    spin_loop();
}
```

- **Free harvesting**: while the sync caller waits, the fire-and-forget commands queued ahead of it complete; `pump_completions` reclaims them, advances the fence high-water mark, and frees slots — a blocking call doubles as free cleanup time for the async machinery.
- **Pop only when `peek_used() == token`**: the used ring is FIFO; this fixes defect 1 of `add_notify_wait_pop`.
- The device writes the response **in place into the caller's buffer** (zero-copy: request/response bytes never touch driver-managed memory).

`request_sync` (ctrl.rs:271) is just `add_sync` + `wait_sync`; the `'a: 'b` lifetime expresses "the output borrow does not outlive the input".

### 6.4 `request_sync_fenced` (ctrl.rs:286) — the deadlock fix

```rust
pub(crate) fn request_sync_fenced(.., fence_id: u64) -> Result<u32> {
    let used_len = self.request_sync(transport, inputs, outputs)?;
    if fence_id > self.completed_fence_id {
        self.completed_fence_id = fence_id;          // popping our own FLAG_FENCE response = fence fired
    }
    Ok(used_len)
}
```

**Why this must exist**: a blocking `submit_3d` carries `FLAG_FENCE`; the device writes the response only **after rendering finishes**. Once `wait_sync` pops that response the fence has factually fired — but if the high-water mark is not advanced here, a subsequent `wait_fence(fence_id)` spins forever on a used entry that will never come. This is a real deadlock the user hit at step 10/15 of the 3D example; the `fenced_sync_request_advances_fence` test locks it. Linux doesn't need this path — its submit never blocks, so fences advance only through `pump_completions`.

---

## 7. Waiting for rendering: `wait_fence` / `fence_completed` (ctrl.rs:533/553)

```rust
pub(crate) fn wait_fence(&mut self, transport: &mut impl Transport, fence_id: u64) -> Result {
    while self.completed_fence_id < fence_id {
        self.notify(transport);                      // ← force a delivery decision first
        self.pump_completions(transport)?;
        if self.completed_fence_id >= fence_id { break; }
        spin_loop();
    }
    Ok(())
}

pub(crate) fn fence_completed(&self, fence_id: u64) -> bool {
    self.completed_fence_id >= fence_id              // NOWAIT probe, O(1)
}
```

Why `notify` must be inside the loop: below the threshold of 8 and away from a full ring, `enqueue` performs **no kick decision at all**, so the fenced command may still be sitting undelivered in the accumulator — waiting without delivering would wait forever. This is the same invariant Linux guarantees by calling `virtio_gpu_notify()` before `virtio_gpu_wait_ioctl` (regression test `wait_fence_delivers_undelivered_fenced_command`). The two APIs map to the blocking/NOWAIT probes of the WAIT ioctl (`dma_fence_wait` / `dma_resv_test_signaled`).

The docs also state the polling trap explicitly: the high-water mark only advances when entries are popped, so a polling consumer **without an IRQ handler must call `pump_completions()` itself** in the poll loop (`ctrl_notify` only delivers, it never reclaims).

---

## 8. The async entry layer: `async_api.rs`

One 2D and one 3D example, showing "same wire protocol as the blocking version, different submission":

```rust
pub fn resource_create_2d_async(&mut self, resource_id: u32, width: u32, height: u32) -> Result {
    self.ctrl.enqueue(&mut self.transport,
        &ResourceCreate2D {
            header: CtrlHeader::with_type(Command::RESOURCE_CREATE_2D),
            resource_id, format: Format::B8G8R8A8UNORM, width, height,
        },
        None,        // no second payload
        0,           // no fence
    )?;
    Ok(())           // the token is dropped: callers need not distinguish ring vs parked
}
```

```rust
pub fn submit_3d_async(&mut self, ctx_id: u32, fence_id: u64, cmds: &[u8]) -> Result {
    self.require_virgl()?;
    let req = CmdSubmit3D {
        header: CtrlHeader::with_fence(Command::SUBMIT_3D, ctx_id, fence_id),  // FLAG_FENCE forced
        size: Self::submit_3d_size(cmds)?,
        _padding: 0,
    };
    self.ctrl.enqueue(&mut self.transport, &req, Some(cmds), fence_id)?;  // payload into the heap box
    Ok(())
}
```

- **`with_fence` forces `VIRTIO_GPU_FLAG_FENCE`**: the device writes the used entry — and advances the high-water mark — only **after rendering finishes**. Without it the used-pop happens when the host decodes+enqueues, and `wait_fence` would report completion before rendering is done — a caller reading back render results would race. Linux fences every EXECBUFFER for the same reason (virtgpu_submit.c's `virtio_gpu_init_submit`).
- Every "don't wait" has a Linux precedent: `set_scanout` ↔ `virtio_gpu_primary_plane_update` doesn't wait; `resource_flush` ↔ `virtio_gpu_cmd_resource_flush` doesn't wait; `ctx_attach_resource` ↔ `virtio_gpu_cmd_ctx_attach_resource` doesn't wait (Mesa creates ~6.9 fresh textures per frame, each attached here — this was the last sync RTT on the steady-state path).
- The 4 support APIs (`ctrl_notify`/`pump_completions`/`wait_fence`/`fence_completed`) expose the §3/§5/§7 machinery as-is — reclamation must be delegated to the consumer (IRQ or polling), the inevitable division of labor in a no-std crate without an interrupt framework.

---

### 9.1 Sync `request()` (mod.rs:328) — the whole zero-copy story

```rust
fn request<Req: IntoBytes + Immutable, Rsp: FromBytes + IntoBytes>(&mut self, req: Req) -> Result<Rsp> {
    let mut response = Rsp::new_zeroed();
    self.ctrl.request_sync(
        &mut self.transport,
        &[req.as_bytes()],                 // request: stack-resident req, borrowed straight onto the ring
        &mut [response.as_mut_bytes()],    // response: on this frame's stack, written in place by the device
    )?;
    Ok(response)
}
```

`req` and `response` both live on this frame's stack; `request_sync` borrows them, puts them on the ring, waits, and pops them back — the bytes never move. This is an advantage over Linux: its input is a user-space pointer and it must `vmemdup_user`-copy. `request_with_extra_response` (capset data) and `request_with_data` (a blob's mem-entry table) are structured the same way.

### 9.2 `cursor_request` (mod.rs:371) — cross-queue ordering

```rust
fn cursor_request<Req: IntoBytes + Immutable>(&mut self, req: Req) -> Result {
    self.ctrl.notify(&mut self.transport);               // flush the ctrl accumulator FIRST
    self.cursor_queue.add_notify_wait_pop(&[req.as_bytes()], &mut [], &mut self.transport)
}
```

The cursor queue and the control queue are two virtqueues with **no mutual ordering guarantee**, yet `UPDATE_CURSOR` references resources created by control commands — without flushing the accumulator, the host could process the cursor command before the resource exists. In Linux this ordering falls out of "one notify per ioctl"; here it is explicit. Note this call site uses **upstream's** `add_notify_wait_pop` — the cursor queue carries only single cursor commands, no token confusion is possible, so the upstream function works within its original constraints.

### 9.3 Blocking `submit_3d` (mod.rs:782)

```rust
let req = CmdSubmit3D { header: CtrlHeader::with_fence(..), size: Self::submit_3d_size(cmds)?, .. };
let mut rsp = [0u8; size_of::<CtrlHeader>()];
self.ctrl.request_sync_fenced(
    &mut self.transport,
    &[req.as_bytes(), cmds],              // the command stream as a second readable buffer, zero-copy
    &mut [&mut rsp],
    fence_id,
)?;
CtrlHeader::read_from_bytes(&rsp).expect(..).check_type(Command::OK_NODATA)
```

`submit_3d_size` rejects non-multiple-of-4 lengths: the host feeds `size/4` dwords to `virgl_renderer_submit_cmd`, so a remainder would be **silently truncated** — the driver rejects it up front with `InvalidParam`. Semantics stay upstream: **return means rendering is complete**; for async see `submit_3d_async`.

### 9.4 `Drop` (mod.rs:884)

Only two `queue_unset` calls (so the device stops touching both rings' DMA areas). **In-flight fire-and-forget commands are not waited for** — an unavoidable consequence of async; the docs state it: before freeing DMA memory the device may still touch, the caller must `wait_fence` on the last submitted fence.

---

## 10. How the tests simulate the device (tail of ctrl.rs, 7 state-machine tests)

Built on the crate's existing `FakeHal`/`FakeTransport`: `State::read_write_queue` processes descriptor chains in FIFO order (read input → write response → mark used), and `State::wait_until_queue_notified` / `poll_queue_notified` observe kicks. std threads play the "asynchronously completing device" while the driver side blocks as usual.

| Test | Invariant it locks |
|---|---|
| `fire_and_forget_commands_are_delivered_without_caller_notify` | Threshold-8 backstop delivery — a never-notify consumer's commands can never starve |
| `notify_delivers_accumulated_batch_early` | One kick delivers a whole batch at a transaction boundary; notify with an empty accumulator is a strict no-op |
| `sync_request_roundtrip_zero_copy` | Sync request bytes flow without driver-managed copies; the response is written in place; fence count untouched |
| `parked_commands_flush_in_fifo_order` | Ring-full parking → FIFO-ordered re-add after reclaim (proved via the fence high-water mark) |
| `fenced_sync_request_advances_fence` | A blocking submit popping its own FLAG_FENCE response must advance the high-water mark (deadlock regression) |
| `wait_fence_delivers_undelivered_fenced_command` | A sub-threshold fenced command is force-delivered by `wait_fence` (deadlock regression) |
| `sync_entry_survives_pump_completions` | A sync entry is never stolen by an (IRQ-style) pump; it is left for its waiter |

---

## 11. Invariant summary (reviewer quick reference)

1. virtqueue FIFO ⇒ submission order = execution order; no per-command fence needed. `flush_pending` at every submission point keeps parked commands ordered first.
2. The fence high-water mark advances only where fenced entries are popped, and there are **exactly two** such places (`pump_completions` + `request_sync_fenced`) — missing one deadlocks; both have regression tests.
3. The vbuf arenas are address-stable for life; `PendingSubmit` carries only the `slot` index.
4. `pending[token].take()` is a claim: same-core IRQ reentrancy cannot double-pop; sync entries have no record and pump breaks at them (the FIFO head cannot be skipped).
5. Every used entry returns exactly one vbuf slot ⇒ every degraded wait makes progress.
6. A sync command is owed a response: unconditional kick + retry on QueueFull; the normal-path kick goes through `should_notify` (identical to upstream).
7. `resource_detach_backing` stays synchronous = the barrier that lets callers free backing memory; `Drop` does not wait for in-flight commands — `wait_fence` before freeing DMA memory.

---

## 12. Performance (A/B experiment)

Controlled experiment, single variable = driver submission model. Control arm `bench-sync` = upstream all-sync + 4 API shims; experiment arm = this commit. Same machine, same time window, same QEMU (virtio-vga-gl + KVM + 1 vCPU), same rootfs (v0.0.12), same app (Weston + glmark2-es2-wayland, 60s per scene):

| glmark2 scene | upstream all-sync | this commit | gain |
|---|---|---|---|
| build use-vbo=false | 207 | 239 | +15% |
| build use-vbo=true | 785 | 969 | **+23%** |
| texture nearest | 762 | 957 | +26% |
| texture linear | 796 | 993 | **+25%** |
| texture mipmap | 798 | 992 | +24% |
| shading gouraud | 797 | 999 | +25% |
| **Average (excl. vbo=false)** | **788** | **982** | **+24.6%** |

Frame time (use-vbo=true) **1.28ms → 1.01ms**. The small gain on `use-vbo=false` is expected: the bottleneck is the app's own guest-CPU vertex path, where submission latency is a small share.

**Where the gain comes from**: ① eliminating 3–5 serial blocking RTTs per frame (the present chain + submit); ② **pipelining (the biggest share)** — `submit_3d` no longer waits for rendering, so guest encoding of frame N+1 overlaps host rendering of frame N, and frame time goes from "sum of both" to "max of both"; ③ kick flattening — one MMIO write at the transaction boundary covers the whole batch (`avail_event` suppresses redundant kicks). On the steady-state frame path the **sync RTT count is 0**; the only remaining wait is the fence — a real GPU dependency that cannot be eliminated, only hidden by the pipeline.


zyc107109102:
Once this is merged and verified on your end, could we get a new crates.io
release? Our project depends on this crate via crates.io, so a release is the
way these changes reach us — happy to help validate either side of it.

zyc107109102:
@qwandor

